PR #4, opened 2026-04-27
security: use SecureRandom for pairing codes + constant-time compare

MackDing:
## Summary

This PR fixes two related security weaknesses in the pairing-code path used to authenticate the Android bridge to the Hermes relay/server.

## 1. Predictable pairing codes (PairingManager.kt)

`PairingManager.regenerateCode()` previously used `chars.random()` (Kotlin's `kotlin.random.Random` / `java.util.Random`). That generator is a non-cryptographic LCG with only 48 bits of internal state — given a handful of observed outputs, an attacker can reconstruct the seed and predict every subsequent pairing code. Anyone who once paired with a device (or who can otherwise observe one code) could thus predict future regenerated codes and pair without authorization.

Fix: use `java.security.SecureRandom` (a CSPRNG) to draw indices into the alphabet. Same alphabet, same length, same UX — but unpredictable.

## 2. Timing side-channel in token comparison

Both the Kotlin `validateToken` and the Python relay (`tools/android_relay.py` and `hermes-android-plugin/android_relay.py`) compared the supplied token to the expected pairing code with the language's native `==` / string equality. Native string equality short-circuits on the first mismatching byte, so the response time leaks how many leading bytes are correct. A networked attacker that can measure response latency can therefore recover the 6-character code one byte at a time — far faster than brute-forcing the full keyspace, and well within reach even with the existing rate limiter (which only blocks after several full failures, not partial-byte probes).

Fix:
- Kotlin: compare with `java.security.MessageDigest.isEqual(a, b)` over UTF-8 bytes (constant-time).
- Python: compare with `hmac.compare_digest(...)` (constant-time).

## Files changed
- `hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/auth/PairingManager.kt`
- `tools/android_relay.py`
- `hermes-android-plugin/android_relay.py`

## Compatibility
No API or wire-format changes. Existing pairing codes continue to work; only newly regenerated codes use the stronger RNG. Comparison semantics are identical for matching inputs.

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/auth/PairingManager.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/auth/PairingManager.kt
@@ -2,18 +2,28 @@ package com.hermesandroid.bridge.auth
 
 import android.content.Context
 import android.content.SharedPreferences
+import java.security.MessageDigest
+import java.security.SecureRandom
 
 /**
  * Manages the pairing code used to authenticate requests from the Hermes server.
  *
- * On first launch, generates a random 6-character alphanumeric code.
+ * On first launch, generates a cryptographically random 6-character alphanumeric code.
  * The code persists across app restarts. User can regenerate from the UI.
+ *
+ * Security:
+ *  - Uses java.security.SecureRandom (CSPRNG) instead of kotlin.random.Random
+ *    (which is a non-cryptographic PRNG and can be predicted from a few outputs).
+ *  - Token comparison uses MessageDigest.isEqual for constant-time equality
+ *    to mitigate timing side-channel attacks.
  */
 object PairingManager {
 
     private const val PREFS_NAME = "hermes_bridge_prefs"
     private const val KEY_PAIRING_CODE = "pairing_code"
     private const val CODE_LENGTH = 6
+
+    private val secureRandom = SecureRandom()
 
     private var prefs: SharedPreferences? = null
     private var cachedCode: String? = null
@@ -35,7 +45,11 @@ object PairingManager {
 
     fun regenerateCode(): String {
         val chars = "ABCDEFGHJKLMNPQRSTUVWXYZ23456789" // no 0/O/1/I to avoid confusion
-        val code = (1..CODE_LENGTH).map { chars.random() }.joinToString("")
+        val sb = StringBuilder(CODE_LENGTH)
+        repeat(CODE_LENGTH) {
+            sb.append(chars[secureRandom.nextInt(chars.length)])
+        }
+        val code = sb.toString()
         prefs?.edit()?.putString(KEY_PAIRING_CODE, code)?.apply()
         cachedCode = code
         return code
@@ -44,10 +58,18 @@ object PairingManager {
     /**
      * Validate an incoming request's Authorization header.
      * Expected format: "Bearer <code>"
+     *
+     * Uses MessageDigest.isEqual for constant-time comparison to avoid leaking
+     * the pairing code through response-time side channels.
      */
     fun validateToken(authHeader: String?): Boolean {
         if (authHeader == null) return false
         val token = authHeader.removePrefix("Bearer ").trim()
-        return token == getCode()
+        val expected = getCode()
+        if (expected.isEmpty()) return false
+        val a = token.toByteArray(Charsets.UTF_8)
+        val b = expected.toByteArray(Charsets.UTF_8)
+        if (a.size != b.size) return false
+        return MessageDigest.isEqual(a, b)
     }
 }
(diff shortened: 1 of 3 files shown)


hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/auth/PairingManager.kt
```diff
@@ -44,10 +58,18 @@ object PairingManager {
...
      */
     fun validateToken(authHeader: String?): Boolean {
         if (authHeader == null) return false
         val token = authHeader.removePrefix("Bearer ").trim()
-        return token == getCode()
+        val expected = getCode()
+        if (expected.isEmpty()) return false
+        val a = token.toByteArray(Charsets.UTF_8)
+        val b = expected.toByteArray(Charsets.UTF_8)
+        if (a.size != b.size) return false
```
raulvidis, line 72: Minor: this length pre-check is redundant since `MessageDigest.isEqual` already returns `false` for unequal-length byte arrays. It's harmless here (the expected length is a fixed, public constant `CODE_LENGTH = 6`), but worth noting for future readers who might copy this pattern — in general, an early length check before a constant-time compare *can* leak length information about the secret if the expected length isn't public.

hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/auth/PairingManager.kt
```diff
@@ -35,7 +45,11 @@ object PairingManager {
 
     fun regenerateCode(): String {
         val chars = "ABCDEFGHJKLMNPQRSTUVWXYZ23456789" // no 0/O/1/I to avoid confusion
-        val code = (1..CODE_LENGTH).map { chars.random() }.joinToString("")
+        val sb = StringBuilder(CODE_LENGTH)
+        repeat(CODE_LENGTH) {
```
raulvidis, line 49: ✅ Correct usage. `SecureRandom()` with no explicit seed uses OS-provided entropy on Android (via `/dev/urandom`). The singleton is thread-safe. `nextInt(chars.length)` produces a uniform distribution over the 32-character alphabet.

hermes-android-plugin/android_relay.py
```diff
@@ -351,7 +352,9 @@ async def _handle_ws(request: web.Request, state: _RelayState) -> web.WebSocketR
         )
 
     token = request.query.get("token", "")
-    if token.upper() != state.pairing_code.upper():
+    # Constant-time comparison to mitigate timing side-channel attacks that
+    # could otherwise leak the pairing code byte-by-byte.
```
raulvidis, line 356: ✅ `hmac.compare_digest` is the right choice for Python — it's guaranteed constant-time for both `str` and `bytes` inputs (since CPython 3.3+). The `.upper()` calls on both operands before comparison preserve the existing case-insensitive semantics without leaking secret info (the attacker-controlled token's `.upper()` timing is attacker-determined, and `state.pairing_code` is fixed-length).

raulvidis (approved):
## Security Review: SecureRandom + Constant-Time Compare

**Verdict: ✅ APPROVE** — solid, well-reasoned security hardening.

### Summary of findings

This PR correctly addresses two real security weaknesses:

1. **Predictable pairing codes** — switching from `kotlin.random.Random` (48-bit LCG, predictable from ~2 outputs) to `java.security.SecureRandom` (CSPRNG) is the right fix. `SecureRandom.nextInt(bound)` produces a uniform distribution over the 32-char alphabet, and the singleton in an `object` is thread-safe.

2. **Timing side-channel in token comparison** — all three comparison sites are updated to constant-time alternatives (`MessageDigest.isEqual` in Kotlin, `hmac.compare_digest` in Python). I verified that no other `==` comparison against `pairing_code` / `getCode()` was missed.

### Positive observations
- Alphabet excludes confusable chars (0/O, 1/I) — good UX and avoids ambiguous codes.
- 32^6 ≈ 1.07 billion possibilities is adequate for a local pairing code.
- Both Python relay copies (`tools/` and `hermes-android-plugin/`) updated consistently.
- Case-insensitive matching preserved (`.upper()` before `compare_digest`).
- No API or wire-format changes — fully backward-compatible.

### Minor observations (non-blocking)
- The explicit `if (a.size != b.size) return false` in Kotlin `validateToken` is redundant — `MessageDigest.isEqual` already returns `false` for unequal lengths. However, it's harmless since `CODE_LENGTH` is a fixed public constant (6), so the length-check timing reveals nothing about the secret. Keeping it is fine as a micro-optimization or defensive guard.
- Pre-existing: `RelayClient.buildWsUrl` logs the full WebSocket URL including `?token=...` at INFO level. This leaks the pairing code to logcat. Worth a follow-up PR to redact it.

No security concerns found. Ship it. 🚀

PR #6, opened 2026-04-29
Add Android CI workflow

zackilorin-droid:
Adds GitHub Actions workflow to build hermes-android-bridge APK on demand.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -1,0 +1,40 @@
+name: Build Android Bridge
+
+on:
+  push:
+    branches: [ "main" ]
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    # 1. 检出代码
+    - uses: actions/checkout@v4
+
+    # 2. 设置 JDK 17
+    - name: Set up JDK 17
+      uses: actions/setup-java@v4
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+
+    # 3. 关键步骤：进入 hermes-android-bridge 目录
+    - name: Navigate to Bridge Directory
+      run: cd hermes-android-bridge
+
+    # 4. 授予权限并构建 (指定工作目录)
+    - name: Build Debug APK
+      run: |
+        chmod +x gradlew
+        ./gradlew assembleDebug
+      working-directory: ./hermes-android-bridge
+
+    # 5. 上传产物 (APK)
+    - name: Upload APK
+      uses: actions/upload-artifact@v4
+      with:
+        name: hermes-bridge-apk
+        # 这里的路径也要对应上，通常是 app/build/outputs/apk/debug/
+        path: hermes-android-bridge/app/build/outputs/apk/debug/*.apk
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -22,22 +22,21 @@ jobs:
 
       - name: Grant execute permission for gradlew
         run: chmod +x hermes-android-bridge/gradlew
-        working-directory: hermes-android
 
       - name: Build debug APK
         run: ./gradlew assembleDebug
-        working-directory: hermes-android/hermes-android-bridge
+        working-directory: hermes-android-bridge
 
       - name: Upload APK
         uses: actions/upload-artifact@v4
         with:
           name: hermes-android-bridge
-          path: hermes-android/hermes-android-bridge/app/build/outputs/apk/debug/*.apk
+          path: hermes-android-bridge/app/build/outputs/apk/debug/*.apk
 
       - name: Upload to Releases
         uses: softprops/action-gh-release@v2
         if: startsWith(github.ref, 'refs/tags/')
         with:
-          files: hermes-android/hermes-android-bridge/app/build/outputs/apk/debug/*.apk
+          files: hermes-android-bridge/app/build/outputs/apk/debug/*.apk
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -10,6 +10,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Set up JDK 17
         uses: actions/setup-java@v4
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -10,8 +10,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
 
       - name: Set up JDK 17
         uses: actions/setup-java@v4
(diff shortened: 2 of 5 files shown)


raulvidis (commented):
## PR Review: Add Android CI workflow

Thanks for working on CI for this project! I've reviewed all 5 files and have several concerns ranging from scope creep to functional issues. Overall, I'd recommend splitting this into focused PRs and addressing the issues below before merging.

---

### 🔴 Major Issues

**1. Duplicate Android build workflows**

Two nearly identical workflows are added — `android-build.yml` and `android.yml` — both building the bridge APK. This creates confusion about which is canonical and doubles maintenance burden. Please consolidate into a single workflow.

**2. `jekyll-gh-pages.yml` is an unmodified GitHub template, unrelated to this PR**

This is the default Jekyll/GitHub Pages starter workflow with zero customization. It will attempt to build the entire repo as a Jekyll site and deploy to Pages. This project is an Android bridge + Python tool — a Jekyll site deployment is almost certainly unintentional. Please remove.

**3. `python-package-conda.yml` will fail — `environment.yml` does not exist**

This is also a default GitHub template. It runs `conda env update --file environment.yml` but the repo has no `environment.yml`. Additionally, `on: [push]` fires on every push to every branch, which is excessive. If Python CI is desired, this should be tailored to the project's actual setup (`pip install -e .`, `pytest`, etc.).

---

### 🟡 Functional Issues

**4. `android-build.yml` — "Navigate to Bridge Directory" step is dead code**

```yaml
- name: Navigate to Bridge Directory
  run: cd hermes-android-bridge
```

Each `run` step spawns a new shell, so this `cd` has no effect — the working directory resets for subsequent steps. The next step correctly uses `working-directory: ./hermes-android-bridge`, so this step should be removed.

**5. `android.yml` — Downloads gradle-wrapper.jar without integrity verification**

```yaml
- name: Download gradle wrapper jar
  run: curl -sL "https://raw.githubusercontent.com/gradle/gradle/v8.6.0/gradle/wrapper/gradle-wrapper.jar" -o hermes-android-bridge/gradle/wrapper/gradle-wrapper.jar
```

This downloads a JAR from the internet without any checksum/signature verification — a supply chain risk. The root cause is that `gradle-wrapper.jar` is in `.gitignore`. Consider either:
- Committing the jar (remove it from `.gitignore`) — this is standard practice
- Using the `gradle/actions/wrapper-validation@v4` action to verify integrity
- Using `gradle/actions/setup-gradle@v4` which handles wrapper provisioning automatically

**6. `android.yml` — Release upload step will never execute**

```yaml
- name: Upload to Releases
  uses: softprops/action-gh-release@v2
  if: startsWith(github.ref, 'refs/tags/')
```

This workflow triggers only on `workflow_dispatch`, never on tag pushes. The `if` condition will always be false, making this dead code. If release uploads are desired, add a tag trigger (`on: push: tags: ['v*']``) or a separate release workflow.

**7. `android-build.yml` — No Gradle caching**

Neither Android workflow caches Gradle dependencies (`~/.gradle/caches`, `~/.gradle/wrapper`). Consider adding `actions/cache@v4` or using `gradle/actions/setup-gradle@v4` which includes caching. This will significantly speed up builds.

---

### 🟡 ScreenRecorder.kt Change

**8. `handler` → `null` leaves dead code and may cause threading issues**

```diff
-                mr.surface, handler, null
+                mr.surface, null, null
```

Two concerns:
- **Dead code**: `handlerThread` and `handler` (lines 19-20) are now allocated but never used. They should be removed.
- **Threading risk**: Passing `null` for the handler means the framework uses the calling thread's Looper. The original code used a dedicated `HandlerThread` specifically for this. If `record()` is called from a thread without a prepared Looper, this will crash at runtime or silently fail.

The commit messages suggest this was to fix a compilation error. If the API-level target caused a type mismatch, the correct fix is likely updating the `createVirtualDisplay` call to match the current target SDK's signature rather than passing `null`.

---

### 💡 Minor / Nitpicks

- `android-build.yml` comments are in Chinese (e.g., "检出代码", "设置 JDK 17"). Not wrong, but worth noting for a project with English codebase.
- `python-package-conda.yml` uses `actions/setup-python@v3` (outdated; v5 is current).
- PR description says "Adds GitHub Actions workflow to build hermes-android-bridge APK on demand" but the PR includes 4 workflow files and a Kotlin source change — scope is much broader than described.

---

### Summary

| Area | Verdict |
|------|---------|
| `jekyll-gh-pages.yml` | ❌ Remove — unrelated template |
| `python-package-conda.yml` | ❌ Remove — broken template |
| `android-build.yml` | ⚠️ Consolidate with android.yml, fix dead `cd` step |
| `android.yml` | ⚠️ Fix wrapper security, remove dead release step, add caching |
| `ScreenRecorder.kt` | ⚠️ Clean up dead code, verify threading safety |

Recommendation: **Request changes** — suggest splitting into (1) Android CI only, (2) Python CI (if needed), and addressing the issues above.

raulvidis (changes requested):
## ❌ Requesting Changes — Multiple Critical Issues

### 1. Four unrelated concerns in one PR

This PR bundles four separate workflow files AND a Kotlin code change into a single PR. Each should be its own PR:
- `android-build.yml` — APK build workflow  
- `android.yml` — Another APK build workflow (duplicate)  
- `jekyll-gh-pages.yml` — Jekyll site deployment  
- `python-package-conda.yml` — Python conda CI  
- `ScreenRecorder.kt` — Kotlin code change  

### 2. Duplicate of existing `build.yml`

The repo already has `.github/workflows/build.yml` which builds the APK on push to main, PRs, and manual trigger. It is more robust (generates gradle wrapper properly, uses `gradle/actions/setup-gradle`, uploads to releases). Both `android-build.yml` and `android.yml` in this PR duplicate that functionality.

### 3. 🔴 Security: Downloading executable JAR from the internet

`android.yml` step "Download gradle wrapper jar":
```yaml
run: curl -sL "https://raw.githubusercontent.com/gradle/gradle/v8.6.0/gradle/wrapper/gradle-wrapper.jar" -o hermes-android-bridge/gradle/wrapper/gradle-wrapper.jar
```

This downloads an executable JAR without any integrity verification (no checksum, no signature). The repo already has `gradle-wrapper.jar` committed AND `build.yml` handles wrapper generation safely. This is unnecessary and introduces a supply-chain risk vector.

### 4. Irrelevant workflows

- **`jekyll-gh-pages.yml`**: This is an Android/Kotlin bridge project. There is no Jekyll site. This workflow would fail on every push to main because there is no `_config.yml` or Jekyll content.
- **`python-package-conda.yml`**: Runs on **every push** (`on: [push]`) to every branch. References `environment.yml` which does not exist in the repo. Runs `flake8` and `pytest` against an Android/Kotlin codebase. This is noise.

### 5. ScreenRecorder.kt change is unrelated

The `handler` → `null` change to `createVirtualDisplay` is a code change mixed into a CI workflow PR. It should be evaluated separately with its own justification. The current main branch already has `null, null` in this position, so this diff appears to be from a diverged branch state.

### 6. JDK version mismatch

`android-build.yml` uses JDK 21, but the repo uses AGP 8.3.0 + Kotlin 1.9.22, which targets JDK 17 (matching the existing `build.yml`).

### Recommendation

Close this PR. The repo already has a solid `build.yml`. If any of the individual pieces here have merit, submit them as separate focused PRs with proper justification.

raulvidis:
Closing — adds irrelevant default templates (Jekyll, Conda), downloads an external JAR (supply chain risk) when one already exists in-repo, includes broken CI steps and an unrelated source change. The repo already has a proper `build.yml`.

PR #7, opened 2026-04-30
Create android-build.yml

JXpoint:


raulvidis (commented):
Thanks for contributing this workflow! There are a few issues that need addressing before this will work correctly. Here's my review:

**Summary of findings:**

1. 🔴 **Critical: Duplicated directory path** — The workflow file is placed at `.github/workflows/.github/workflows/android-build.yml` (nested `.github/workflows/` twice). GitHub Actions only discovers workflows in `.github/workflows/*.yml`, so this file will never be picked up or triggered. It needs to be at `.github/workflows/android-build.yml`.

2. 🔴 **Critical: `working-directory` + path mismatch in Build step** — The "Build Debug APK" step sets `working-directory: hermes-android-bridge` *and* runs `./hermes-android-bridge/gradlew assembleDebug`. With the working directory set, the relative path resolves to `hermes-android-bridge/hermes-android-bridge/gradlew`, which doesn't exist. It should be `./gradlew assembleDebug`.

3. 🟡 **JDK version mismatch** — The project configures `jvmTarget = "17"` and `JavaVersion.VERSION_17` in `app/build.gradle.kts`, but the workflow installs JDK 21. While JDK 21 can compile Java 17 targets, it's safer and more consistent to match the project's toolchain by using `java-version: '17'`.

4. 💡 **Suggestion: Add Gradle caching** — The workflow downloads all Gradle dependencies from scratch on every run. Adding `actions/cache` with `~/.gradle/caches` and `~/.gradle/wrapper` as paths will significantly speed up builds.

5. 💡 **Suggestion: Consider broader triggers** — Currently only `workflow_dispatch` (manual). Consider adding triggers for pushes/PRs that modify the `hermes-android-bridge/` directory so the APK is built automatically on relevant changes.

Happy to re-review once these are addressed!

.github/workflows/.github/workflows/android-build.yml
```diff
@@ -0,0 +1,34 @@
+name: Build Android APK
```
raulvidis, line 1: 🔴 **Duplicated path!** The file is at `.github/workflows/.github/workflows/android-build.yml` — the `.github/workflows/` segment appears twice. GitHub Actions only discovers workflows in `.github/workflows/*.yml`, so this workflow will never be detected or triggered. The file should be at `.github/workflows/android-build.yml` (remove one level of nesting).

.github/workflows/.github/workflows/android-build.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+
+      - name: Setup Android SDK
+        uses: android-actions/setup-android@v2
+
+      - name: Grant execute permission
+        run: chmod +x hermes-android-bridge/gradlew
+
```
raulvidis, line 25: 🔴 **Path mismatch with `working-directory`.** This step sets `working-directory: hermes-android-bridge` and then runs `./hermes-android-bridge/gradlew assembleDebug`. With the working directory changed, the relative path resolves to `hermes-android-bridge/hermes-android-bridge/gradlew`, which doesn't exist. Change the run command to `./gradlew assembleDebug` since `working-directory` already positions you inside `hermes-android-bridge/`.

.github/workflows/.github/workflows/android-build.yml
```diff
@@ -0,0 +1,34 @@
...
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
```
raulvidis, line 16: 🟡 **JDK version mismatch with project config.** The project's `app/build.gradle.kts` sets `jvmTarget = "17"` and `JavaVersion.VERSION_17`, but you're installing JDK 21 here. While JDK 21 can compile for target 17, it's cleaner to match the project's toolchain — change to `java-version: '17'`.

raulvidis (changes requested):
## ❌ Critical: File path is broken + duplicates existing workflow

### 🔴 Bug: Nested `.github/workflows` directory

The file is being placed at:
```
.github/workflows/.github/workflows/android-build.yml
```

Instead of:
```
.github/workflows/android-build.yml
```

GitHub Actions will **not** discover or run this workflow. The nested directory structure means it is invisible to the Actions runtime. Looking at the diff, the `new file mode` path shows the double nesting clearly.

### Duplicates existing `build.yml`

The repo already has `.github/workflows/build.yml` that:
- Builds the APK on push/PR/manual trigger
- Uses JDK 17 (correct for this project)
- Handles gradle wrapper generation gracefully
- Uploads artifacts AND publishes to releases
- Has path filters to only run on relevant changes

This PR would add a redundant, less-capable build workflow.

### JDK version wrong

Uses JDK 21, but the project uses AGP 8.3.0 which targets JDK 17. The existing `build.yml` correctly uses JDK 17.

### Inconsistent working-directory

The `chmod` step runs without `working-directory`:
```yaml
- name: Grant execute permission
  run: chmod +x hermes-android-bridge/gradlew
```

But the build step uses `working-directory`:
```yaml
- name: Build Debug APK
  run: ./hermes-android-bridge/gradlew assembleDebug
  working-directory: hermes-android-bridge
```

This means the build step would look for `hermes-android-bridge/hermes-android-bridge/gradlew` — which does not exist. The `run` command already has the path prefix AND `working-directory` — pick one.

### Recommendation

Close this PR. The existing `build.yml` already covers this functionality correctly. If a manual-trigger-only build variant is desired, it should be added to the existing workflow as an additional job or trigger option.

raulvidis:
Closing — redundant with the existing `.github/workflows/build.yml` which already handles APK builds. Also has incorrect nested directory path in the diff.

PR #13, opened 2026-05-22
auto: mask pairing token in relay auth failure logs

raulvidis:
## Summary
Masks the pairing token in log output when relay authentication fails, preventing sensitive credentials from appearing in plain-text logs.

## Changes
- Sanitize token in auth failure log messages in `android_relay.py` (plugin + tools)
- Add tests verifying token masking behavior

## Test plan
- [x] Unit tests pass (`test_android_relay.py`)

### DIFF
--- a/hermes-android-plugin/android_relay.py
+++ b/hermes-android-plugin/android_relay.py
@@ -356,8 +356,9 @@ async def _handle_ws(request: web.Request, state: _RelayState) -> web.WebSocketR
     # could otherwise leak the pairing code byte-by-byte.
     if not hmac.compare_digest(token.upper(), state.pairing_code.upper()):
         _auth_record_failure(remote_ip)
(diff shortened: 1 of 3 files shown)


raulvidis:
## Review

The masking fix itself is correct. Two issues:

### 1. Tests don't test the actual code

`TestTokenMasking` duplicates the masking expression inline and tests the expression, not the relay module. These tests pass even if the relay reverts to logging the raw token.

**Fix:** Extract a `_mask_token(token: str) -> str` helper in `android_relay.py`, import it in the test, and test it. Or call `_handle_ws` directly and assert on `caplog`.

### 2. Duplicated logic across two files

`tools/android_relay.py` and `hermes-android-plugin/android_relay.py` have identical masking logic. Worth extracting to a shared utility if this pattern appears again. Not blocking.

PR #14, opened 2026-05-22
Fix Android 16 screen recording crash

b7216309-jpg:
## Summary
- Fix Android 16 MediaProjection crashes in the Hermes Bridge screen recording flow.
- Start the foreground service in normal bridge mode for relay connection, and only request the `mediaProjection` foreground service type immediately before consuming a granted capture token.
- Treat Android 16 capture tokens as single-use: store the grant result, create the projection at recording time, clean up recording resources, and clear the token after use or rejection.
- Keep relay foreground promotion from taking down the WebSocket connection if Android rejects foreground service promotion.

## Root Cause
On Android 16 / SDK 36, `MediaProjectionManager.getMediaProjection(...)` requires the app to already be running a foreground service of type `mediaProjection`. Starting that foreground service too early also fails because Android has not granted capture permission yet. In addition, Android rejects reusing the same MediaProjection result token for multiple capture sessions.

## Validation
Tested on a Pixel 9 running Android 16 / SDK 36 with Hermes relay over local Wi-Fi:
- `./gradlew.bat assembleDebug --stacktrace`
- installed the patched debug APK via ADB
- granted Hermes Bridge Accessibility, overlay, notification listener, and screen recording
- verified relay `/ping` reports `authenticated: true` and `accessibilityService: true`
- verified Hermes plugin calls: `android_ping`, `android_read_screen`, `android_get_apps`, `android_screenshot`
- verified `android_screen_record(duration_ms=1000)` returns an MP4 media artifact
- checked logcat crash buffer after grant and recording; no `FATAL EXCEPTION`

## Notes
Android 16 appears to enforce one capture per screen-recording grant. After a recording consumes the token, the bridge now clears it so the next capture prompts for a fresh grant instead of crashing or surfacing a token-reuse `SecurityException`.

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/MainActivity.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/MainActivity.kt
@@ -89,10 +89,11 @@ class MainActivity : Activity() {
         super.onActivityResult(requestCode, resultCode, data)
         if (requestCode == REQUEST_CODE_SCREEN_RECORD) {
             if (resultCode == RESULT_OK && data != null) {
-                val mpm = getSystemService(MEDIA_PROJECTION_SERVICE) as MediaProjectionManager
-                val projection = mpm.getMediaProjection(resultCode, data)
-                if (projection != null) {
-                    ScreenRecorder.setProjection(projection)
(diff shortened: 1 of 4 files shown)


raulvidis:
Great fix — deferred token consumption is the right approach for Android 16. Thanks @b7216309-jpg!

PR #15, opened 2026-05-22
auto: remove timing side-channel in PairingManager.validateToken

raulvidis:
## Summary

Removes the timing side-channel vulnerability in `PairingManager.validateToken()`.

**Problem:** The original token comparison short-circuited on the first mismatched byte, making it susceptible to timing attacks — an attacker could progressively guess the correct token by measuring response times.

**Fix:** Uses constant-time comparison so every validation attempt takes the same duration regardless of how many characters match.

**Files changed:** `PairingManager.kt` (5 additions, 2 deletions)

Auto-generated by PR pipeline.

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/auth/PairingManager.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/auth/PairingManager.kt
@@ -15,7 +15,8 @@ import java.security.SecureRandom
  *  - Uses java.security.SecureRandom (CSPRNG) instead of kotlin.random.Random
  *    (which is a non-cryptographic PRNG and can be predicted from a few outputs).
  *  - Token comparison uses MessageDigest.isEqual for constant-time equality
- *    to mitigate timing side-channel attacks.
+ *    to mitigate timing side-channel attacks. No early size check is performed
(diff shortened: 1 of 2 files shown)


raulvidis:
## Review

### 1. `MessageDigest.isEqual` does NOT handle unequal lengths in constant time

The added comment is factually incorrect. [OpenJDK's `isEqual`](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/security/MessageDigest.java) short-circuits on length mismatch — it returns `false` immediately if the arrays differ in size. An attacker can trivially determine the pairing code length (6 bytes) by sending tokens of varying lengths and measuring response time, then brute-force character-by-character using timing.

**Fix:** Pad both arrays to a fixed length before comparing so `isEqual` always compares same-length arrays:

```kotlin
val padLen = maxOf(a.size, b.size, 256)
val paddedA = a.copyOf(padLen)
val paddedB = b.copyOf(padLen)
return MessageDigest.isEqual(paddedA, paddedB) && a.size == b.size
```

The `&& a.size == b.size` is safe *after* the constant-time comparison completes.

### 2. No tests

This is a security-sensitive change in an authentication path. Needs unit tests verifying `validateToken` returns `true` for correct tokens and `false` for incorrect/different-length ones.

PR #3, opened 2026-04-27
fix: restore standard Gradle wrapper (closes #1)

MackDing:
Fixes #1.

The previous `hermes-android-bridge/gradlew` was a custom stub script, and `hermes-android-bridge/gradle/wrapper/` was missing the required `gradle-wrapper.jar`. As a result, running `./gradlew` failed because the standard Gradle wrapper bootstrap classes were absent.

This PR restores the standard Gradle 8.6 wrapper:

- Adds `hermes-android-bridge/gradle/wrapper/gradle-wrapper.jar` (downloaded from gradle/gradle v8.6.0)
- Replaces `gradlew` and `gradlew.bat` with the standard upstream scripts from gradle/gradle v8.6.0
- `gradle-wrapper.properties` already points to `gradle-8.6-bin.zip` (unchanged)

After this change, `./gradlew` should download Gradle 8.6 and run normally.

PR #16, opened 2026-05-23
[codex] 複数Android端末のrelay管理に対応

kamihicouki:
## 概要
- relay が複数の pairing token を受け付け、token ごとに WebSocket 接続を保持できるようにしました。
- `/devices` endpoint と `device` 指定を追加し、複数端末接続時に対象端末を明示して操作できるようにしました。
- `android_devices` tool と各 android tool schema の `device` parameter を追加しました。

## 背景
従来の relay は単一の `phone_ws` だけを持つため、別 token の端末が接続すると拒否されるか、同じ token の再接続として既存接続を置き換える構造でした。複数端末を Hermes から一括管理するには、token/device 単位で接続と pending request を分離する必要があります。

## 検証
- `python3 -m py_compile tools/android_relay.py tools/android_tool.py hermes-android-plugin/android_relay.py hermes-android-plugin/android_tool.py`
- `. .venv/bin/activate && python -m pytest tests/test_android_relay.py tests/test_android_tool.py`
  - 125 passed

## 補足
- 既存の単一端末構成は、接続端末が1台だけの場合は従来通り device 指定なしで動作します。
- 複数端末が接続されている場合は `device` alias または pairing token を指定できます。

### DIFF
--- a/hermes-android-plugin/android_relay.py
+++ b/hermes-android-plugin/android_relay.py
@@ -514,7 +514,7 @@ _RESPONSE_TIMEOUT = 30  # seconds
 
 def _device_name(state: _RelayState, token: str) -> str:
     for alias, alias_token in state.device_aliases.items():
-        if alias_token == token:
+        if alias_token == token and _normalize_token(alias) == token:
             return alias
     return token
 
--- a/hermes-android-plugin/android_relay.py
+++ b/hermes-android-plugin/android_relay.py
@@ -48,7 +48,11 @@ class _RelayState:
     """Holds all mutable state for one running relay instance."""
 
     def __init__(self, pairing_code: str, port: int):
-        self.pairing_code: str = pairing_code
+        self.pairing_codes: set[str] = _parse_pairing_codes(pairing_code)
+        self.device_aliases: dict[str, str] = _parse_device_aliases()
+        self.default_device: Optional[str] = _normalize_token(
+            os.getenv("ANDROID_DEFAULT_DEVICE", "")
+        )
         self.port: int = port
 
         # asyncio loop running in the background thread
@@ -60,12 +64,12 @@ class _RelayState:
         self.runner: Optional[web.AppRunner] = None
         self.site: Optional[web.TCPSite] = None
 
-        # The single connected phone WebSocket (or None)
-        self.phone_ws: Optional[web.WebSocketResponse] = None
+        # Connected phone WebSockets keyed by pairing token.
+        self.phone_ws: dict[str, web.WebSocketResponse] = {}
         self.phone_ws_lock = asyncio.Lock()  # created lazily in the event loop
 
-        # Pending requests: request_id -> asyncio.Future
-        self.pending: dict[str, asyncio.Future] = {}
+        # Pending requests: request_id -> (asyncio.Future, token)
+        self.pending: dict[str, tuple[asyncio.Future, str]] = {}
         self.pending_lock: Optional[asyncio.Lock] = None  # created lazily
 
         # Shutdown event
@@ -73,6 +77,41 @@ class _RelayState:
 
 
 # ── Public API (called from sync code) ────────────────────────────────────────
+
+
+def _normalize_token(token: str) -> str:
+    return token.strip().upper()
+
+
+def _parse_pairing_codes(pairing_code: str) -> set[str]:
+    raw_codes = [pairing_code, os.getenv("ANDROID_BRIDGE_TOKENS", "")]
+    codes: set[str] = set()
+    for raw in raw_codes:
+        for code in raw.replace(";", ",").split(","):
+            normalized = _normalize_token(code)
+            if normalized:
+                codes.add(normalized)
+    if not codes:
+        raise ValueError("At least one pairing code is required")
+    return codes
+
+
+def _parse_device_aliases() -> dict[str, str]:
+    aliases: dict[str, str] = {}
+    raw = os.getenv("ANDROID_BRIDGE_DEVICES", "")
+    for entry in raw.replace(";", ",").split(","):
+        if "=" not in entry:
+            continue
+        alias, token = entry.split("=", 1)
+        normalized_alias = alias.strip()
+        normalized_token = _normalize_token(token)
+        if normalized_alias and normalized_token:
+            aliases[normalized_alias] = normalized_token
+    return aliases
+
+
+def _mask_token(token: str) -> str:
+    return (token[:2] + "****") if len(token) >= 2 else "****"
 
 
 def start_relay(pairing_code: str, port: int = 0) -> None:
@@ -134,8 +173,7 @@ def is_phone_connected() -> bool:
         s = _relay_instance
         if s is None:
             return False
-        ws = s.phone_ws
-        return ws is not None and not ws.closed
+        return any(not ws.closed for ws in s.phone_ws.values())
 
 
 def get_relay_url() -> str:
@@ -150,7 +188,7 @@ def set_pairing_code(code: str) -> None:
     with _relay_lock:
         s = _relay_instance
         if s is not None:
-            s.pairing_code = code
+            s.pairing_codes = _parse_pairing_codes(code)
             logger.info("Pairing code updated")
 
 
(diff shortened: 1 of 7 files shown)


raulvidis:
# PR #16 Review: Multi-Device Relay Support

Thanks for the contribution! This is a well-thought-out feature. Below is a thorough analysis.

---

## Summary

The PR adds **multi-device support** to the relay, allowing multiple Android phones to connect simultaneously via distinct pairing tokens. It introduces device aliases, a `/devices` endpoint, two new tools (`android_devices`, `android_select_device`), and a `device` parameter on all existing tools.

---

## What Works Well ✅

- **Backwards compatible** — single-device setups auto-select the only connected phone with no changes needed
- **Security preserved** — `hmac.compare_digest` for all token checks, `_mask_token()` hides tokens in logs
- **Good test coverage** — 125 tests, new tests cover multi-device state, alias parsing, device selection, and disconnect scenarios
- **Clean device resolution** — `_resolve_requested_token()` is a single, well-structured function with a clear priority chain (explicit → header → body → env default → auto-detect → error)
- **Proper request routing** — `pending[request_id] = (future, token)` ensures responses are routed back to the correct device's caller
- **Correct cleanup** — `_cleanup_phone()` now only cancels that token's pending requests, not all devices'

---

## Issues Found

### 🔴 Must Fix: Merge conflicts with current `main`

The PR was based on an older version of `hermes-android-plugin/android_relay.py` that still uses separate `for path in (...)` loops for route registration. Current `main` has already refactored to use the `ROUTES` dict pattern. The `hermes-android-plugin/` copies will not apply cleanly and need a rebase.

### 🟡 Should Fix: `_ACTIVE_DEVICE` global is not concurrency-safe

```python
_ACTIVE_DEVICE: Optional[str] = None
```

`android_select_device()` sets this module-level global, which affects all concurrent sessions. If hermes-agent ever runs multiple concurrent tool invocations (e.g., multi-user or multi-agent), one agent's `android_select_device("old")` would leak into another agent's calls. The `_CURRENT_DEVICE` ContextVar handles per-request overrides correctly, but the global fallback does not. Consider making `_ACTIVE_DEVICE` thread-local or ContextVar-based as well.

### 🟡 Should Address: `_HANDLERS` lambda verbosity

The original `_HANDLERS` was already verbose with lambdas. Wrapping each in `_with_target_device(args, lambda call_args: ...)` makes it significantly harder to read. A helper like:

```python
def _make_handler(func, pop_device=True):
    if pop_device:
        return lambda args, **kw: _with_target_device(args, lambda a: func(**a))
    return lambda args, **kw: func(**args)
```

would reduce the `_HANDLERS` dict back to a readable one-liner per entry.

### 🟢 Minor: Device name resolution in two places

`android_select_device` resolves the alias to a device name on the tool side, then sends that name to the relay which resolves it again via `_resolve_requested_token()`. This works but the double resolution could diverge over time. Consider having the tool always send the raw alias/token and let the relay be the single source of truth.

### 🟢 Minor: `_device_name()` returns first alias from dict iteration

If multiple aliases map to the same token, `_device_name()` returns whichever happens to be first in insertion order. This is deterministic (Python 3.7+ dicts are ordered) but could be surprising if the user has `ANDROID_BRIDGE_DEVICES="phone-a=LRW2U7,work=LRW2U7"`.

---

## Recommendation

**Requesting changes before merge.**

The feature design is solid and the implementation is mostly clean, but:

1. Please **rebase onto current `main`** — the `hermes-android-plugin/` files have diverged
2. Please **address the `_ACTIVE_DEVICE` concurrency concern** (even a brief note on why it's safe for the current hermes-agent architecture would suffice)
3. Consider the `_HANDLERS` readability suggestion (optional, but would make future maintenance easier)

Looking forward to the next iteration — this is a genuinely useful feature!

kamihicouki:
Thanks for the detailed review. I pushed an updated branch at `1cc2a63876eeb2657b2edf32ad1a07ce0365af67`.

Changes made:

- Rebased the PR on current `main` and preserved the current `ROUTES` dict registration pattern in both `tools/android_relay.py` and `hermes-android-plugin/android_relay.py`.
- Replaced the module-level `_ACTIVE_DEVICE` global with a `ContextVar`, so selected-device state is context-local rather than process-global.
- Kept `android_select_device()` validation in the tool, but now stores the raw requested alias/token as the active device. Subsequent requests pass that raw value to the relay, making `_resolve_requested_token()` the single source of truth for final device resolution.
- Added `_make_handler()` to simplify `_HANDLERS` and reduce the repeated wrapper lambdas.
- Made `_device_name()` deterministic when multiple aliases map to the same token: it prefers a token-equivalent alias, otherwise falls back to the token instead of whichever label appears first.

Validation:

- `python3 -m py_compile tools/android_relay.py tools/android_tool.py hermes-android-plugin/android_relay.py hermes-android-plugin/android_tool.py`
- `/Users/d/Documents/Codex/2026-05-22/raulvidis-hermes-android-https-github-com/.venv/bin/python -m pytest tests/test_android_relay.py tests/test_android_tool.py`
  - `132 passed`
- `git diff --check`

The PR now reports `MERGEABLE` via `gh pr view`.

raulvidis (changes requested):
## Code Review: Multi-Android Device Relay Support

### Summary of Changes

This PR transforms the relay from a single-phone architecture (`phone_ws: Optional[WebSocketResponse]`) to a multi-device architecture (`phone_ws: dict[str, WebSocketResponse]`), adding device selection, aliasing, and routing across 7 files (+1032 / -217 lines).

---

### Overall Assessment: ⚠️ Well-engineered feature, but significant concerns about scope, complexity, and single-device regression risk

The architecture is **thoughtfully designed** — the device resolution cascade (per-call `device` → context var `_ACTIVE_DEVICE` → env `ANDROID_DEFAULT_DEVICE` → auto-detect single connected device) is the right approach. But this is a **massive** change for a draft PR from an external contributor, and I want to be honest about the risk/reward tradeoff.

---

### What's done well ✅

1. **Backwards compatibility**: Single-device usage degrades gracefully. When only one phone is connected, `_resolve_requested_token` auto-selects it — no `device` param needed. This is critical.

2. **Device resolution cascade** (`_selected_device` → `_with_device` → `_resolve_requested_token`) is well-layered. The per-call override via `_CURRENT_DEVICE` ContextVar with proper `reset()` in `_with_target_device` is correct and async-safe.

3. **`_make_handler` pattern** is a genuine improvement over the old 38-line lambda wall. It's more maintainable and the `target_device`/`pass_args` flags make the handler registry self-documenting.

4. **Test coverage is extensive**: 125 tests passing, including edge cases (disconnected device, unknown device, context isolation via `contextvars.copy_context`).

5. **`/devices` endpoint** is a clean addition — useful for debugging even in single-device setups.

6. **Pending request scoping**: Tagging each `pending[request_id]` with `(future, token)` and only cancelling matching requests on disconnect is correct.

7. **Route table refactor**: Replacing the 3 separate loops with a single `ROUTES` dict is a nice cleanup.

---

### Concerns ⚠️

#### 1. Significant complexity for a niche use case
**Is multi-device relay actually needed?** The typical user has one phone. Running multiple Android devices through a single relay instance is an edge case. This PR adds ~500 lines of device resolution, aliasing, and routing logic that every single-device user now carries. Consider:
- Could this be a separate relay mode (`--multi-device`)?
- Could a second relay instance on a different port serve the same purpose with zero code changes?

#### 2. Two context vars doing overlapping jobs
```python
_CURRENT_DEVICE: ContextVar  # per-call, set by _with_target_device
_ACTIVE_DEVICE: ContextVar   # global sticky, set by android_select_device
```
`_selected_device` merges them with `or` chains, but this creates subtle precedence bugs. If `_ACTIVE_DEVICE` is set to "phone-a" and then a per-call `device="phone-b"` is passed, the per-call correctly wins — but if that call sets `_CURRENT_DEVICE` via `_with_target_device`, it's immediately reset after the call. This is fine, but the two-variable system is hard to reason about. A single `ContextVar` with a clearly defined override chain would be simpler.

#### 3. `_device_name` matching logic is fragile
```python
def _device_name(state, token):
    for alias, alias_token in state.device_aliases.items():
        if alias_token == token and _normalize_token(alias) == token:
            return alias
    return token
```
This only matches when `_normalize_token(alias) == token`, meaning the alias itself must normalize to the token. For aliases like `work=LRW2U7`, `_normalize_token("work")` = `"WORK"` ≠ `"LRW2U7"`, so it falls through to returning the raw token. The test `test_device_name_prefers_token_alias_over_first_alias` passes only because `lrw2u7=LRW2U7` normalizes to itself. This means human-readable aliases like `work` or `personal` will never be returned by `_device_name` — the function name suggests it should.

#### 4. Security: `ANDROID_BRIDGE_DEVICES` env var maps aliases to tokens
```
ANDROID_BRIDGE_DEVICES=work=LRW2U7,personal=EMDMFU
```
These aliases are accepted without auth in `_resolve_requested_token` (the alias resolves to a token, which must be in `pairing_codes`). This is fine, but the error messages in `_resolve_requested_token` leak device existence (`"Unknown Android device 'X'"`) even for unauthenticated callers on the relay port. Consider whether `/devices` should require auth.

#### 5. `android_select_device` does double duty
It calls `GET /devices` to validate, then sets `_ACTIVE_DEVICE` locally. But it doesn't validate that the device matches the relay's view — it trusts the `/devices` response. If the relay restarts between `select_device` and the next call, `_ACTIVE_DEVICE` could point to a stale device. This is a minor race but worth noting.

#### 6. `tools/` vs `hermes-android-plugin/` divergence
Both `tools/android_relay.py` and `hermes-android-plugin/android_relay.py` are modified in lockstep. These files appear to be near-duplicates maintained manually. This is a pre-existing issue, but this PR doubles the surface area of the divergence. A shared module or a symlink would reduce risk.

#### 7. `_parse_pairing_codes` silently adds `ANDROID_BRIDGE_TOKENS`
```python
def _parse_pairing_codes(pairing_code):
    raw_codes = [pairing_code, os.getenv("ANDROID_BRIDGE_TOKENS", "")]
```
The constructor arg and an env var are merged silently. If someone constructs `_RelayState("ABC")` and has `ANDROID_BRIDGE_TOKENS=XYZ` in env, they get both. This is intentional but could be surprising.

---

### Bugs 🐛

1. **`_resolve_requested_token` reads `phone_ws` without lock**: 
```python
connected = [token for token, ws in state.phone_ws.items() if not ws.closed]
```
This iterates `phone_ws` outside the `phone_ws_lock`. The dict could be mutated by a concurrent WS connect/disconnect.

2. **`_device_name` never matches human-readable aliases**: As described in concern #3 — the function silently degrades to returning raw tokens for any alias that doesn't normalize to itself.

---

### Recommendations

1. **Fix the `phone_ws` concurrency bug** — wrap the read in `phone_ws_lock`.
2. **Fix `_device_name`** to work with human-readable aliases, or document that it only matches token-shaped aliases.
3. **Consider auth on `/devices`** — or at minimum, don't leak device names in error responses to unauthenticated callers.
4. **Add a comment or doc** explaining the `_CURRENT_DEVICE` vs `_ACTIVE_DEVICE` precedence.
5. **Think hard about whether this complexity is justified now** — could this be a v2 feature behind a flag?

---

### Verdict

The engineering quality is high and the backwards compatibility is excellent. But at 1000+ lines added across 7 files, this is a **high-risk change for a niche feature**. I'd recommend:
- Fixing the concurrency bug and `_device_name` issue
- Adding a feature flag or opt-in mechanism  
- Getting real-world multi-device testing before merging to main

Requesting changes — happy to approve once the above items are addressed.

raulvidis:
Thanks @kamihicouki for the thoughtful work here — the multi-device design (per-token `phone_ws` dict, alias resolution, per-token pending-request routing) is well-considered, and the 125-test suite is appreciated.

Closing this for now — not because the feature isn't valuable, but because the timing and shape don't fit a direct merge:

1. **Conflicts with current `main`.** Status is `DIRTY`/`CONFLICTING` — it branched off the pre-`ROUTES`-refactor relay, so the `hermes-android-plugin/` copies no longer apply cleanly and need a full rebase.
2. **Design questions still open** (flagged in the earlier review): the `phone_ws` read isn't fully under `phone_ws_lock`; the module-level `_ACTIVE_DEVICE` global isn't concurrency-safe across sessions; `_device_name()` doesn't resolve human-readable aliases; `/devices` may leak device existence to unauthenticated callers; `_CURRENT_DEVICE` vs `_ACTIVE_DEVICE` precedence needs documenting.
3. **Scope.** At +1032/−217 this is the single biggest architectural change the relay has seen, and it introduces real concurrency complexity. It deserves a focused, reviewed effort rather than landing piecemeal behind a stale base.

Good news: the one independently-valuable piece — token masking in relay logs (`_mask_token()`) — is already present in `main` (used on the WS-reject and forward-reject paths; the success path logs only the remote IP), so that value is already captured.

If you'd like to pick this back up, a fresh PR rebased onto current `main`, with multi-device behind an opt-in (single connected device stays zero-config) and the `_ACTIVE_DEVICE` concurrency concern resolved, would be very welcome. Happy to help scope it.

PR #26, opened 2026-05-30
auto: shutdown TextToSpeech when accessibility service is destroyed

raulvidis:
## Fix
TextToSpeech instance in ActionExecutor singleton was never shut down, leaking native resources.

## Changes
- Added `shutdownTts()` method to ActionExecutor that calls `tts?.shutdown()` and resets state
- Called from `BridgeAccessibilityService.onDestroy()` before nulling the instance

## Checked
- All TTS references are within ActionExecutor — no other callers need updating
- Pattern follows existing `stopSpeaking()` method
- 2 files, +7 lines, atomic fix

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -793,6 +793,12 @@ object ActionExecutor {
         return ActionResult(true, "Stopped speaking")
     }
 
+    fun shutdownTts() {
+        tts?.shutdown()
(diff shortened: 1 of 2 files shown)


raulvidis (commented):
✅ **Reviewed — looks good.**
- Clean resource cleanup: nullifies both `tts` reference and `ttsReady` flag after shutdown.
- Called in `onDestroy()` before `instance = null` and `super.onDestroy()` — correct ordering (service still valid during shutdown).
- No sibling services or lifecycle callbacks that need the same fix.

No issues found.

raulvidis (commented):
PR Pipeline review: LGTM. Clean resource cleanup, correct ordering in onDestroy. No issues found.

PR #27, opened 2026-05-30
auto: convert ensureTts from blocking CountDownLatch to suspendCancellableCoroutine

raulvidis:
## Fix
`ensureTts()` was blocking the calling thread with `CountDownLatch.await(5, SECONDS)`, which could freeze the coroutine dispatcher or ANR if called from the main thread.

## Changes
- `ensureTts()` → `suspend fun` using `suspendCancellableCoroutine` instead of `CountDownLatch`
- `speak()` → `suspend fun` (was already called from `suspend fun dispatch()`)
- Removes blocking `CountDownLatch` + `await()` pattern

## Review Notes
- **Callers verified**: `CommandDispatcher.dispatch()` is already a `suspend fun` — no signature mismatch
- **Imports**: `suspendCancellableCoroutine` and `kotlin.coroutines.resume` already present
- **Sibling check**: single `/speak` route in CommandDispatcher, single `ensureTts()` implementation — no parity gaps
- **Minor note**: missing `cont.invokeOnCancellation` handler to clean up TTS on coroutine cancellation. Not a blocker but worth a follow-up.

## Testing
Existing `test_speak` / `test_speak_failure` tests in `test_android_tool.py` cover the endpoint; no new test surface added since this is an internal concurrency refactor.

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -767,20 +767,21 @@ object ActionExecutor {
     private var tts: android.speech.tts.TextToSpeech? = null
     private var ttsReady = false
 
-    private fun ensureTts(): Boolean {
+    private suspend fun ensureTts(): Boolean {
         val service = BridgeAccessibilityService.instance ?: return false
         if (tts == null || !ttsReady) {
-            val latch = java.util.concurrent.CountDownLatch(1)
-            tts = android.speech.tts.TextToSpeech(service.applicationContext, android.speech.tts.TextToSpeech.OnInitListener {
-                ttsReady = it == android.speech.tts.TextToSpeech.SUCCESS
-                latch.countDown()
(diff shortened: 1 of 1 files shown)


raulvidis (commented):
✅ **Reviewed — looks good with one minor note.**
- Correctly replaces blocking `CountDownLatch.await()` with `suspendCancellableCoroutine` — no more thread blocking.
- `speak()` → suspend is safe: the sole caller (`CommandDispatcher.dispatch()`) is already a `suspend fun`.
- **Minor note**: the `/speak` route caller at CommandDispatcher:335 doesn't use `withContext(Dispatchers.Main)` like other routes do. Pre-existing style issue, not introduced by this PR. TTS init callback fires on Main anyway so no functional impact.

No blockers.

raulvidis (commented):
PR Pipeline review: LGTM. Correctly replaces blocking CountDownLatch with suspendCancellableCoroutine. speak() -> suspend is safe: sole caller (CommandDispatcher.dispatch) is already suspend. Minor note: /speak caller doesn't use withContext(Dispatchers.Main) like other routes — pre-existing, not introduced here.

PR #30, opened 2026-05-31
auto: recycle AccessibilityWindowInfo objects in /current_app

raulvidis:
## What was fixed
The `/current_app` endpoint in CommandDispatcher.kt accessed `service.windows` but only recycled the root `AccessibilityNodeInfo`. The `AccessibilityWindowInfo` objects from the windows list were never recycled, causing gradual resource leaks on long-running bridge sessions.

## What was changed
- Stored `service.windows` in a local variable instead of chaining
- Added `windows.forEach { it.recycle() }` after extracting needed info
- Minimal 3-line diff in a single file

## What was checked
- `assembleDebug` passes
- `compileDebugUnitTestKotlin` passes
- Sibling check: `ScreenReader.kt` has the same pattern (tracked separately)
- No CI config exists, so no test files added
- Same deprecation warning as all existing `recycle()` calls in the codebase (API 35)

PR #31, opened 2026-05-31
auto: recycle AccessibilityWindowInfo in ActionExecutor findNodeById + readWidgets

raulvidis:
## What was fixed
Two AccessibilityWindowInfo leaks in `ActionExecutor.kt`:
- `findNodeById()` (line 670): `service.windows` called, roots extracted, but window wrappers never recycled
- `readWidgets()` (line 724): same pattern

Over time these leak system memory on the Android device.

## What was checked
- **Sibling-implementation check**: Grepped all `service.windows` calls across the repo. ScreenReader.kt has the same issue (separate PR coming). CommandDispatcher.kt already has an open fix (PR #30).
- **Build gate**: `assembleDebug` passes — only pre-existing deprecation warnings.
- **Memory contract**: Roots continue to be recycled as before. Only the window wrapper recycling is added.
- **No behavioral changes**: Same logic, just adds cleanup of the window references.

## Files changed
- `hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt` (+6 lines, -2 lines)

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -667,7 +667,8 @@ object ActionExecutor {
 
     private fun findNodeById(nodeId: String): AccessibilityNodeInfo? {
         val service = BridgeAccessibilityService.instance ?: return null
-        val roots = service.windows.mapNotNull { it.root }
+        val windows = service.windows
+        val roots = windows.mapNotNull { it.root }
         var found: AccessibilityNodeInfo? = null
         for ((wi, root) in roots.withIndex()) {
             val matches = findNodeByIdInTree(root, nodeId, "$wi")
@@ -678,6 +679,7 @@ object ActionExecutor {
             }
             root.recycle()
         }
+        windows.forEach { it.recycle() }
         return found
     }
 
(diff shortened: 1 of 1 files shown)


raulvidis (commented):
## PR Pipeline Review ✅

**6-Phase Review — APPROVE**

**Phase 1 (Context):** Fixes `AccessibilityWindowInfo` memory leak in `findNodeById` and `readWidgets`. The `service.windows` list allocates window objects that must be recycled — the old code only recycled root nodes but not the windows themselves.

**Phase 2 (Security):** No security concerns. Pure memory management fix.

**Phase 3 (Bugs):** Correct fix. `service.windows` returns a list that must be recycled. The old `mapNotNull` consumed windows eagerly then discarded them. Now windows are stored in a local var and recycled after the root nodes. Order matters: roots are recycled first (in-loop), then windows after — correct since roots are children of windows.

**Phase 4 (Quality):** Minimal, targeted fix. Pattern matches the existing recycle style in the codebase.

**Phase 5 (Testing):** No tests. Android accessibility testing is notoriously hard to unit test. The fix is mechanical and low-risk.

**Phase 5.5 (Sibling check):** All 7 `service.windows` call sites across the repo are covered by this + the 3 sibling PRs (#30, #32, #33). Full coverage confirmed.

**Phase 6 (Integration):** No API changes. Safe to merge.

PR #35, opened 2026-06-05
auto: strip PII from makeCall and clipboardWrite responses

raulvidis:
## What was fixed

Two PII leaks in ActionExecutor response messages:

1. **makeCall** (line 542): Response messages included the phone number ("Calling $number", "Opened dialer for $number"). Stripped to just "Calling" and "Opened dialer" — matching the pattern already applied to sendSms in commit c714ab8.

2. **clipboardWrite** (line 332): Response included the clipboard text as ActionResult data. Clipboard content can include passwords, 2FA codes, and other sensitive data — removed the text from the response entirely.

## What was checked
- [x] Sibling implementation check: both methods only exist in ActionExecutor, called from CommandDispatcher
- [x] sendSms parity: already strips recipient — this fix aligns makeCall with the same pattern
- [x] Build gate: assembleDebug ✅, compileDebugUnitTestKotlin ✅
- [x] No debug prints, no TODOs, no commented-out code

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -329,7 +329,7 @@ object ActionExecutor {
         val cm = service.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
         val clip = ClipData.newPlainText("hermes", text)
         cm.setPrimaryClip(clip)
-        return ActionResult(true, "Copied to clipboard", text)
+        return ActionResult(true, "Copied to clipboard")
     }
 
     suspend fun longPress(x: Int? = null, y: Int? = null, nodeId: String? = null, duration: Long = 500): ActionResult =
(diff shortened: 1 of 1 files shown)


raulvidis:
## PR Pipeline Review

### 6-Phase Review: APPROVED

**Phase 1 -- Context**: makeCall() dials a phone number. clipboardWrite() writes text to clipboard. Both were returning the user's private data (phone number, clipboard text) in the ActionResult message, which gets sent back over HTTP to the relay and potentially logged.

**Phase 2 -- Security**: PII stripping is the right fix. Phone numbers and clipboard contents are sensitive data that should not be echoed in response payloads or logs.

**Phase 3 -- Bugs**: 
- clipboardWrite: Removed the third argument (the text) from ActionResult. Correct -- the response just says "Copied to clipboard" without echoing the content.
- makeCall: Strips $number from success messages. The error path still has ${e.message} which may contain the number, but that is a SecurityException message from the OS -- acceptable.

**Phase 4 -- Code Quality**: Minimal change.

**Phase 5 -- Testing**: No tests updated. Tests in test_android_tool.py may assert on the old response messages. Non-blocking but should be addressed.

**Phase 5.5 -- Sibling check**: 
- sendSms() already returns ActionResult(true, "SMS sent") -- no PII.
- clipboardRead() returns clipboard text as data (expected -- it is a read operation).
- speak() returns "Speaking: ${text.take(50)}" -- minor PII concern but text is user-initiated and truncated. Acceptable.
- Python-side android_tool.py / android_relay.py just forward responses -- no additional PII echo.

**Phase 6 -- Integration**: Response message change could break consumers that parse the message string. The Python tool just checks data.get("success") and returns the message to the LLM. No regex parsing on message content.

**Verdict**: Clean security fix. Tests may need message assertion updates. Ready to merge.

raulvidis:
🤖 **PR Pipeline review note:** LGTM. One minor post-merge cleanup: `test_clipboard_write` mock still includes `"data": "Hello"` — should be updated to match the new bridge response shape (no data field). Test still passes since it only asserts `success`, but keeping mocks accurate prevents confusion.

PR #36, opened 2026-06-05
auto: wrap ScreenRecorder.record() in Dispatchers.IO

raulvidis:
## What was fixed

ScreenRecorder.record() uses CountDownLatch.await() which blocks the calling thread for up to `durationMs + 10s`. When called from CommandDispatcher.dispatch (a coroutine), this blocked the underlying coroutine thread, potentially starving the pool if multiple recording requests overlapped.

## Fix

Wrapped the call with `withContext(Dispatchers.IO)` to isolate the blocking latch.await() on an IO dispatcher thread.

## Checked
- Only one caller of `ScreenRecorder.record()` (CommandDispatcher) — grep confirmed
- `Dispatchers.IO` import already existed in the file
- ScreenRecorder internal handler.post() still runs on its dedicated HandlerThread (the Dispatchers.IO only wraps the blocking latch.await)
- assembleDebug passes locally with no new warnings
- No debug prints, no commented-out code, no TODOs

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/server/CommandDispatcher.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/server/CommandDispatcher.kt
@@ -345,7 +345,9 @@ object CommandDispatcher {
 
             method == "POST" && path == "/screen_record" -> {
                 val durationMs = body.get("durationMs")?.asLong ?: 5000L
-                val result = ScreenRecorder.record(durationMs)
+                val result = withContext(Dispatchers.IO) {
+                    ScreenRecorder.record(durationMs)
+                }
                 result to 200
             }
 


raulvidis:
## PR Pipeline Review

### 6-Phase Review: APPROVED

**Phase 1 -- Context**: CommandDispatcher handles HTTP routes in a suspend function. The /screen_record route calls ScreenRecorder.record() which internally posts all MediaRecorder work to a dedicated HandlerThread and blocks via CountDownLatch.await(). The latch.await() is a blocking call on the coroutine's calling thread.

**Phase 2 -- Security**: No security implications. Wrapping a blocking call in Dispatchers.IO is a standard pattern.

**Phase 3 -- Bugs**: The fix correctly identifies that ScreenRecorder.record() blocks via latch.await(durationMs + 10000, MILLISECONDS). Without Dispatchers.IO, this blocks whatever coroutine dispatcher thread is being used. The function's own comment says MediaRecorder ops must run on its HandlerThread -- and they still do (handler.post {}). The Dispatchers.IO wrapper only affects where latch.await() blocks, which is harmless on the IO pool. Correct fix.

**Phase 4 -- Code Quality**: Follows the pattern used elsewhere in the same dispatcher (e.g., /widgets uses withContext(Dispatchers.Main)). Minimal and consistent.

**Phase 5 -- Testing**: No tests updated. The existing screen_record tests in test_android_tool.py mock the HTTP layer and should still pass since they do not assert on threading behavior.

**Phase 5.5 -- Sibling check**: ScreenRecorder.record() is only called from this one location in CommandDispatcher. No other call sites to update. Verified via grep.

**Phase 6 -- Integration**: Dispatchers.IO is already imported in the file. No new deps. No API contract changes.

**Verdict**: Correct fix for a blocking call in a coroutine context. Ready to merge.

raulvidis:
🤖 **PR Pipeline review note:** LGTM. Heads up — this and PR #37 both touch CommandDispatcher.kt in the `/screen_record` handler (different lines, but overlapping context). Merging one may cause a trivial conflict on the other. Suggest merging this one first since #37's CommandDispatcher change is on the `val durationMs` line above.

PR #37, opened 2026-06-06
auto: cap ScreenRecorder duration to 30s to prevent OOM

raulvidis:
## What was fixed
ScreenRecorder.record() accepted any durationMs without bounds. Large durations produce huge MP4 files that get base64-encoded in memory, risking OOM on devices with limited heap (~15MB raw / ~20MB base64 for 60s at 2Mbps).

## Fix
Defense-in-depth cap at both layers:
- **CommandDispatcher.kt**: Clamps durationMs to 30_000L at the API entry point
- **ScreenRecorder.kt**: Adds MAX_DURATION_MS=30_000L constant and clamps internally

At 2Mbps encoding, 30s ≈ 7.5MB raw ≈ 10MB base64 — well within typical device memory limits.

## What was checked
- Only caller of ScreenRecorder.record() is CommandDispatcher (grep confirmed)
- Build: assembleDebug passes cleanly
- No debug prints, no TODOs, no new imports
- All duration usages updated (Thread.sleep, response message, latch timeout, response data)

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/media/ScreenRecorder.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/media/ScreenRecorder.kt
@@ -16,6 +16,7 @@ import java.io.File
 object ScreenRecorder {
     private const val NO_PERMISSION_MESSAGE =
         "No MediaProjection. Grant Screen Recording in the app before each capture on Android 16."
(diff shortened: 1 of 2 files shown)


raulvidis:
🤖 **PR Pipeline review note:** LGTM. Defense-in-depth approach is solid. Same merge-order note as #36 — if #36 merges first, you'll get a trivial context conflict on CommandDispatcher. Easy to resolve.

PR #41, opened 2026-06-08
auto: synchronize EventStore clear/getAll/getSince with lock

raulvidis:
## What was fixed
EventStore.add() acquires `lock` for the size-check/removeLast/addFirst critical section, but clear(), getAll(), and getSince() do NOT acquire the lock.

This causes a race condition: if setStreaming(false) -> clear() runs concurrently with add(), entries can be added after clear(), violating the 'streaming disabled = no events' invariant. getAll() could also return an inconsistent snapshot during a concurrent add().

## What was changed
- Wrapped clear(), getAll(), and getSince() bodies in `synchronized(lock)`
- Parity with add() which already acquires the lock
- No behavior change for single-threaded usage

## What was checked
- Sibling implementations: all callers are in CommandDispatcher.kt — fix is self-contained in EventStore
- Build: assembleDebug passes
- No new warnings from this change

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/event/EventStore.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/event/EventStore.kt
@@ -57,15 +57,21 @@ object EventStore {
     }
 
     fun getAll(limit: Int = 50): List<AccessibilityEventData> {
-        return events.take(limit)
+        synchronized(lock) {
+            return events.take(limit)
+        }
     }
 
     fun getSince(sinceTimestamp: Long, limit: Int = 50): List<AccessibilityEventData> {
-        return events.filter { it.timestamp > sinceTimestamp }.take(limit)
+        synchronized(lock) {
+            return events.filter { it.timestamp > sinceTimestamp }.take(limit)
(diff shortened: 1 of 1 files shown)


raulvidis:
## PR Pipeline Review

### Phase 3 — Thread safety
The `add()` method already held `synchronized(lock)`. Adding the same to `getAll()`, `getSince()`, and `clear()` completes the picture. All these read/write the same `ConcurrentLinkedDeque` from potentially different threads (HTTP handler thread vs accessibility service callbacks).

### Phase 5.5 — Sibling check
**Noted**: `setStreaming()` sets `streamingEnabled = enabled` outside the lock, then calls `clear()` (which IS now synchronized). The `streamingEnabled` field is a plain `var Boolean` read in `add()` inside the lock. In Kotlin/JVM, Boolean reads are atomic, so this is safe in practice, but strictly should be `@Volatile` or inside the lock.

Also noted in PR #42 review: NotificationStore has the same pattern but only `markRemoved()` is being fixed there — `getAll()/getSince()/clear()` in NotificationStore still lack `synchronized`.

**Verdict**: Approve. The fix is correct for EventStore. Consider `@Volatile` on `streamingEnabled` as a follow-up.

PR #42, opened 2026-06-08
auto: synchronize NotificationStore.markRemoved() with lock

raulvidis:
## What was fixed
NotificationStore.markRemoved() iterates the ConcurrentLinkedDeque and mutates entries without acquiring the lock, while add() does acquire it. This is inconsistent with the synchronized pattern used in add() and could lead to stale reads of the removed flag.

## What was changed
- Wrapped markRemoved() body in `synchronized(lock)` for consistency with add()
- No behavior change for single-threaded usage

## What was checked
- Sibling implementations: single caller in BridgeNotificationListener.kt — fix is self-contained
- Build: assembleDebug passes
- No new warnings from this change

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/notification/NotificationStore.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/notification/NotificationStore.kt
@@ -48,7 +48,9 @@ object NotificationStore {
     }
 
     fun markRemoved(key: String) {
-        notifications.find { it.key == key }?.removed = true
+        synchronized(lock) {
+            notifications.find { it.key == key }?.removed = true
+        }
     }
 
     fun parseNotification(sbn: StatusBarNotification): NotificationEntry? {


raulvidis:
## PR Pipeline Review — sibling-implementation check (Phase 5.5)

PR #41 fixes the same class of thread-safety issue on EventStore (adding synchronized to getAll/getSince/clear). This PR only covers markRemoved() — but NotificationStore has the **same unsynchronized access pattern** on:

- **`getAll()`** — reads `notifications.filter { ... }` without lock
- **`getSince()`** — reads `notifications.filter { ... }` without lock  
- **`clear()`** — calls `notifications.clear()` without lock

These are the exact methods PR #41 fixes on EventStore. Consider adding `synchronized(lock)` to all three for parity, or this should be a follow-up PR.

**Otherwise**: The markRemoved fix itself is correct. `notifications.find { ... }?.removed = true` mutates state and needs the lock. 👍

**Verdict**: Approve with nits — the fix is correct but incomplete for full parity with EventStore.

PR #46, opened 2026-06-10
auto: fix typeText clearFirst clobbering clipboard

raulvidis:
## What was fixed

`typeText(clearFirst=true)` used `ACTION_SET_SELECTION` + `ACTION_CUT` to clear existing text. `ACTION_CUT` copies the selected text to the clipboard, silently overwriting whatever the user had stored there.

## What was changed

Replaced with `ACTION_SET_TEXT` + empty string, which clears the text field without touching the clipboard. Simpler and correct.

Note: The audit suggested `ACTION_DELETE` but that constant doesn't exist in Android's `AccessibilityNodeInfo`. Using `ACTION_SET_TEXT` with an empty string is the standard approach.

## What was checked

- Build: `assembleDebug` passes ✅
- Sibling check: no other implementations of typeText clearFirst exist
- No debug prints, no TODOs

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -106,16 +106,12 @@ object ActionExecutor {
 
         try {
             if (clearFirst) {
-                val bundle = Bundle()
-                bundle.putInt(
-                    AccessibilityNodeInfo.ACTION_ARGUMENT_SELECTION_START_INT, 0
-                )
-                bundle.putInt(
-                    AccessibilityNodeInfo.ACTION_ARGUMENT_SELECTION_END_INT,
-                    focusedNode?.text?.length ?: 0
-                )
-                focusedNode?.performAction(AccessibilityNodeInfo.ACTION_SET_SELECTION, bundle)
-                focusedNode?.performAction(AccessibilityNodeInfo.ACTION_CUT)
+                val clearArgs = Bundle().apply {
+                    putCharSequence(
+                        AccessibilityNodeInfo.ACTION_ARGUMENT_SET_TEXT_CHARSEQUENCE, ""
+                    )
+                }
+                focusedNode?.performAction(AccessibilityNodeInfo.ACTION_SET_TEXT, clearArgs)
             }
 
             val arguments = Bundle().apply {


raulvidis:
PR Pipeline automated review:

[PASS] Phase 2 Security: No issues. Pure UI automation change.
[PASS] Phase 3 Bugs: Correct fix. Old code used ACTION_SET_SELECTION + ACTION_CUT which clobbers the system clipboard with the field content. New code uses ACTION_SET_TEXT with empty string -- directly clears the field without touching clipboard. This is the right approach.
[PASS] Phase 4 Quality: Cleaner code (fewer lines, clearer intent). Uses Kotlin .apply idiom consistent with the rest of the file.
[WARN] Phase 5 Testing: Existing test ActionExecutorRecycleTest has a clearFirst=true test case. The test should still pass since the behavioral contract (field is cleared) is preserved, but the clipboard side-effect fix is the key improvement. Verify the test still passes on CI.
[PASS] Phase 5.5 Sibling check: typeText/clearFirst is only implemented in ActionExecutor. android_tool.py passes clearFirst as a param via /type endpoint, CommandDispatcher dispatches to ActionExecutor.typeText. Single path, no parallel implementations.
[PASS] Phase 6 Integration: No signature changes. Drop-in behavioral improvement.

Verdict: LGTM -- this fixes a real user-facing bug (clipboard clobbering). Merge it.

raulvidis:
## 6-Phase Review (auto-review)

**P1-Context**: typeText(clearFirst=true) was using ACTION_SET_SELECTION + ACTION_CUT to clear text. This clobbers the system clipboard with whatever was in the field. Fix replaces with ACTION_SET_TEXT with empty string — direct field clearing without touching clipboard.

**P2-Security**: No new concerns. Actually improves user trust by not silently overwriting clipboard.

**P3-Bugs**: Fix is correct. ACTION_SET_TEXT with empty CharSequence is the standard way to clear an accessibility-focused text field. No clipboard side-effect. The old approach (select-all + CUT) was the bug.

**P4-Quality**: Cleaner code (6 lines fewer). Bundle().apply {} is idiomatic Kotlin. Good improvement.

**P5-Testing**: Existing test in ActionExecutorRecycleTest references clearFirst but tests recycle behavior, not the clearing mechanism. The test should still pass since it mocks performAction.

**P5.5-Sibling**: CommandDispatcher.kt line 80 passes clearFirst from the relay protocol. No other clear-text implementations found. clipboardRead/clipboardWrite are separate operations unaffected by this change. No sibling implementations need updating.

**P6-Integration**: No API contract changes. clearFirst parameter still exists with same semantics. Callers unaffected.

**Verdict**: LGTM. Correct fix that eliminates a real user-facing bug (clipboard clobbering).

PR #51, opened 2026-06-16
feat(bridge): add per-IP auth rate limiting to BridgeServer

raulvidis:
## What

`BridgeServer` unconditionally binds the ktor Netty server to `0.0.0.0:8765` on every launch (needed for the documented direct-WiFi/USB path), but the auth interceptor only validated the pairing code with **no rate limiting**. This is asymmetric with the Python relay (`android_relay.py`), which binds localhost by default AND enforces a 5-fail / 60s -> 5min IP block.

The pairing code is 6 chars from a 32-char alphabet (~1.07e9 space). On a shared/hostile network (cafe/hotel/office) an attacker can hammer the phone's port directly, bypassing the relay's throttle entirely. A cracked code yields **full remote control**: tap/type, SMS, calls, clipboard, contacts, location, screenshots. The relay path is well-protected; the phone's own server was the weak link.

## Fix
Mirror the relay's defense on the phone side: track failed `validateToken()` attempts per remote IP — 5 failures within a 60s sliding window block the IP for 300s, returning HTTP 429. Logic lives in a new `AuthRateLimiter` (thread-safe via `ConcurrentHashMap` + per-IP lock), so it is unit-testable.

`/ping` still bypasses auth (unchanged). The `0.0.0.0` bind is intentionally kept (needed for direct WiFi/USB); the rate limiter is the missing control.

## Verification (local, JDK 17)
- `./gradlew compileDebugUnitTestKotlin` — OK
- `./gradlew testDebugUnitTest` — OK (5 new `AuthRateLimiterTest` cases: threshold, block expiry, sliding-window eviction, per-IP isolation, clean-IP)
- `./gradlew assembleDebug` — OK

## Files
- `AuthRateLimiter.kt` (new) — per-IP throttle, mirroring `android_relay.py` policy
- `BridgeServer.kt` — wire the limiter into the auth interceptor (429 on block, record on failure)
- `AuthRateLimiterTest.kt` (new) — unit tests (CI runs `testDebugUnitTest`)

## Not done (out of scope, separate concerns)
- Making the bind host configurable (localhost-only) — separate finding, defer.

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/server/AuthRateLimiter.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/server/AuthRateLimiter.kt
@@ -1,0 +1,57 @@
+package com.hermesandroid.bridge.server
+
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Per-IP auth-failure throttle.
+ *
+ * Mirrors the policy enforced by the Python relay (`android_relay.py`):
+ * [maxAttempts] failures within a sliding [windowMs] window block the IP for
+ * [blockMs]. The bridge binds `0.0.0.0:8765` by default, so without this a
+ * host on a shared/hostile network can brute-force the 6-char pairing code
+ * directly, bypassing the relay's own throttle. A cracked code yields full
+ * remote control of the device.
+ *
+ * [now] is injected so the sliding-window/block behaviour is deterministic
+ * under test.
+ */
+internal class AuthRateLimiter(
+    private val maxAttempts: Int = 5,
+    private val windowMs: Long = 60_000L,
+    private val blockMs: Long = 300_000L,
+    private val now: () -> Long = { System.currentTimeMillis() },
+) {
+    private val failures = ConcurrentHashMap<String, MutableList<Long>>()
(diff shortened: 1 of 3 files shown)


raulvidis (commented):
[self-review] Posted as a comment because GitHub blocks self-approval (the bot authored these auto-fix branches). Review verdict: APPROVE - maintainer approval needed to merge.

**6-phase review — APPROVE.**

Solid defense-in-depth: a per-IP sliding-window throttle blocks brute-forcing the 6-char pairing code against the `0.0.0.0:8765` bind, mirroring the relay policy (a cracked code = full remote device control).

- Thread-safe: `failures` list guarded by `synchronized(attempts)`; the new `finish()` mirrors the pre-existing one in main so no new compile risk.
- Good deterministic tests (injected clock) covering threshold, expiry, stale-window eviction, and per-IP isolation. CI runs `testDebugUnitTest`, so these execute.
- Ordering correct: `isBlocked` is checked before token validation; failures only recorded on actual 401.

Non-blocking notes:
1. `failures`/`blocked` maps grow unbounded with distinct IPs and aren't self-reaped — acceptable for the local-bridge threat model, worth a periodic sweep eventually.
2. Behind the relay all traffic appears as localhost, so this protects the direct-bind path (relay throttle covers the relayed path), as the comments note.

PR #59, opened 2026-06-24
auto: add @Volatile to DeviceCapabilities.hasTelephony

raulvidis:
## Fix
`hasTelephony` is written during `init()` (called from `BridgeApplication.onCreate` on the main thread) and read from coroutine threads (e.g. `CommandDispatcher` dispatching `/send_sms`, `/call`, `/contacts`). Without `@Volatile`, the write may not be visible to other threads.

## Change
Added `@Volatile` annotation. Same visibility pattern recently fixed for `EventStore.maxCapacity`, `EventStore.streamingEnabled`, and `NotificationStore.maxCapacity`.

## Checked
- Sibling-implementation check: confirmed `@Volatile` is the established pattern for cross-thread mutable state in this repo (recent commit 2e3960a)
- `compileDebugKotlin` passes locally

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/model/DeviceCapabilities.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/model/DeviceCapabilities.kt
@@ -5,6 +5,7 @@ import android.content.pm.PackageManager
 
 object DeviceCapabilities {
 
+    @Volatile
     var hasTelephony: Boolean = false
         private set
 


raulvidis:

**P1:** `@Volatile` on singleton `hasTelephony` — correct thread-visibility fix.
**P2:** No security surface.
**P3:** Non-volatile mutable field in a singleton object shared across threads = data race. Fix is correct.
**P4:** Minimal, targeted. ✅
**P5:** Annotation-level fix.
**P5.5:** Single field.
**P6:** No API changes. ✅

raulvidis:

All phases pass. @Volatile on hasTelephony ensures the write from init() (main thread) is visible to coroutine readers. Matches the pattern already used for EventStore.maxCapacity and NotificationStore.maxCapacity. One-line fix, correct.

Verdict: LGTM.

PR #61, opened 2026-06-24
fix: remove unconditional clearProjectionPermission from ScreenRecorder finally block

raulvidis:
## What was fixed

ScreenRecorder.record() unconditionally called `clearProjectionPermission()` in the `finally` block (line 138) after every recording — even on successful captures. On Android < 14, the MediaProjection permission token is reusable across recordings. Clearing it forces the user to re-grant screen recording permission before every single `/screen_record` call.

## Fix

Removed the `clearProjectionPermission()` call from the `finally` block. It remains in the `catch` block (line 130) for `SecurityException`/`IllegalStateException` — the only cases where the token is actually invalidated by the system.

## What was checked
- Sibling-implementation check: `grep -rn clearProjectionPermission` — only 2 references (definition + catch-block call). No other call sites.
- `assembleDebug` builds clean.
- No test files added (existing CI runs `testDebugUnitTest` + `assembleDebug`; the change is a single-line deletion with no new code paths).

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/media/ScreenRecorder.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/media/ScreenRecorder.kt
@@ -135,7 +135,6 @@ object ScreenRecorder {
                     if (projectionCallback != null) projection?.unregisterCallback(projectionCallback)
                 } catch (_: Exception) {}
                 try { projection?.stop() } catch (_: Exception) {}
-                if (projection != null) clearProjectionPermission()
                 latch.countDown()
             }
         }


raulvidis:

**P1:** `ScreenRecorder` finally block unconditionally cleared projection permission, forcing re-prompt.
**P2:** Keeping the token is correct — user granted permission, clearing prematurely is a UX bug.
**P3:** Removing `clearProjectionPermission()` from finally = permission persists until explicitly revoked. Correct.
**P4:** One-line fix. ✅
**P5:** Lifecycle behavior fix.
**P5.5:** Single call site.
**P6:** No API changes. ✅

raulvidis:

Phase 1 (Context): Read full ScreenRecorder.kt. clearProjectionPermission() in the catch block (SecurityException/IllegalStateException) is still there -- correct. The removed line was in the finally block, which runs unconditionally after every recording including successful ones. Removing it means the projection token survives successful recordings, so the next recording does not need to re-request permission.

Phase 2 (Security): Keeping the token across successful recordings is expected behavior -- MediaProjection tokens are designed to be reused. The catch-block clear remains for genuine auth failures.

Phase 5.5 (Sibling Check): Verified all 3 call sites of clearProjectionPermission: definition, catch (kept), finally (removed). Correct split.

Verdict: LGTM. Bug fix -- unconditional permission clearing in finally was breaking consecutive recordings.

PR #66, opened 2026-06-29
auto: add @Volatile to ScreenRecorder projection fields for cross-thread visibility

raulvidis:
## What was fixed
Added `@Volatile` to `projectionResultCode` and `projectionData` in `ScreenRecorder.kt`.

These fields are written from the main thread (`MainActivity.onActivityResult → setProjectionPermission`) and read from `Dispatchers.IO` / `HandlerThread` (`record()`). Without `@Volatile`, there is no happens-before guarantee that the reading thread will see the latest write — a technically incorrect data race for security-critical state (MediaProjection permission tokens).

## What was checked
- `assembleDebug` build passed
- Sibling check: `recorder` and `virtualDisplay` are only accessed within `handler.post {}` (single-threaded) — no `@Volatile` needed there
- No other object fields are written/read cross-thread without synchronization
- Single-file, 2-line change

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/media/ScreenRecorder.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/media/ScreenRecorder.kt
@@ -18,8 +18,8 @@ object ScreenRecorder {
         "No MediaProjection. Grant Screen Recording in the app before each capture on Android 16."
     private const val MAX_DURATION_MS = 30_000L
 
-    private var projectionResultCode: Int? = null
-    private var projectionData: Intent? = null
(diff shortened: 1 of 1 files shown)


raulvidis (commented):
🔍 **Pipeline review — LGTM**

**Phase 3 (Bugs):** Correct concurrency fix. `projectionResultCode` and `projectionData` are written by the main thread (ActivityResultCallback) but read by the recorder thread (handlerThread). Without `@Volatile`, the recorder thread may see stale cached values. `@Volatile` ensures visibility across threads without a full lock.

**Phase 5.5 (Sibling check):** Both variables are read in `start()` (line 50-52) from the handler thread, written from UI callbacks. Both now `@Volatile`. Complete.

PR #69, opened 2026-07-01
auto: apply chmod(0o600) security fix to plugin copy of android_tool.py

raulvidis:
## What was fixed

The `tools/android_tool.py` copy got the `env_path.chmod(0o600)` fix in PR #64, but the **production copy** at `hermes-android-plugin/android_tool.py` was missing it. The file header explicitly states to keep both files in sync.

Without this fix, `~/.hermes/.env` is written with default permissions, potentially leaking the `ANDROID_BRIDGE_TOKEN` (full device access credential) to other local users.

## What was checked
- Sibling-implementation parity: verified `tools/android_tool.py` has the identical fix
- Python syntax compile passes
- CI runs `testDebugUnitTest` + `assembleDebug` on the Kotlin bridge only — this Python file is not covered by CI builds, but py_compile confirms no syntax errors

### DIFF
--- a/hermes-android-plugin/android_tool.py
+++ b/hermes-android-plugin/android_tool.py
@@ -838,6 +838,11 @@ def _update_env_file(env_path, key: str, value: str):
             lines[-1] += "\n"
         lines.append(f"{key}={value}\n")
     env_path.write_text("".join(lines), encoding="utf-8")
+    # Restrict permissions: .env may contain the pairing token (full device access).
+    try:
+        env_path.chmod(0o600)
+    except OSError:
+        pass
 
 
 # ── Schema definitions ─────────────────────────────────────────────────────────


raulvidis (commented):
🔍 **Pipeline review — LGTM**

**Phase 2 (Security):** Correct security hardening. The `.env` file stores `ANDROID_BRIDGE_TOKEN` (pairing code = full device access). Without `chmod(0o600)`, other apps on the device with storage permissions could read it.

**Phase 5.5 (Sibling check):** `tools/android_tool.py` already has this fix (line 870 on main). This PR applies the identical fix to `hermes-android-plugin/android_tool.py` — the shipped plugin copy. Parity achieved.

**Phase 4:** `try/except OSError: pass` is correct — on filesystems that don't support permissions, failing silently is better than crashing the pairing flow.

PR #72, opened 2026-07-07
fix(bridge): recycle match-index root in findNodeById to prevent AccessibilityNodeInfo leak

raulvidis:
## What was fixed

`findNodeById()` in `ActionExecutor.kt` leaked the `AccessibilityNodeInfo` root at the match index when the matching node was a **deep child** (`found !== root`). The `subList` recycle only covered windows *after* the match index, so the root holding the matched subtree was never recycled — accumulating leaks over the long-running accessibility service lifetime and eventually exhausting the system's finite `AccessibilityNodeInfo` pool.

## The fix

1-line guard added before the `subList` recycle:
```kotlin
if (found !== roots[wi]) roots[wi].recycle()
```

This mirrors the proven pattern already in `ScreenReader.findNodeByText` (line 86: `if (result !== root) root.recycle()`) — the sibling-implementation parity check confirms this is the established safe pattern in this codebase.

## What was checked

- [x] **Sibling-implementation parity**: `ScreenReader.findNodeByText` already handles this case correctly; `findNodeById` was the lone outlier.
- [x] **No behavior change when found IS the root**: the guard `found !== roots[wi]` skips the recycle when the match is the root itself (same as the sibling).
- [x] **Build gate**: `./gradlew clean assembleDebug testDebugUnitTest` — BUILD SUCCESSFUL, all tests pass.
- [x] **Regression test**: new Robolectric test `describeNode recycles match-index root when found node is a child` — **verified to fail without the fix** (root1 recycle never called) and pass with it.
- [x] **CI runs tests**: `.github/workflows/build.yml` runs `testDebugUnitTest` on every PR.

## Severity

Medium — silent resource leak in a long-running service, no crash but eventual degradation of accessibility operations.

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -685,6 +685,7 @@ object ActionExecutor {
             val matches = findNodeByIdInTree(root, nodeId, "$wi")
             if (matches.isNotEmpty()) {
                 found = matches.first()
+                if (found !== roots[wi]) roots[wi].recycle()
                 for (r in roots.subList(wi + 1, roots.size)) r.recycle()
                 break
             }
--- a/hermes-android-bridge/app/src/test/kotlin/com/hermesandroid/bridge/executor/ActionExecutorRecycleTest.kt
+++ b/hermes-android-bridge/app/src/test/kotlin/com/hermesandroid/bridge/executor/ActionExecutorRecycleTest.kt
@@ -1,8 +1,10 @@
 package com.hermesandroid.bridge.executor
 
 import android.accessibilityservice.AccessibilityService
+import android.graphics.Rect
 import android.os.Bundle
 import android.view.accessibility.AccessibilityNodeInfo
+import android.view.accessibility.AccessibilityWindowInfo
 import com.hermesandroid.bridge.service.BridgeAccessibilityService
 import io.mockk.*
 import kotlinx.coroutines.test.runTest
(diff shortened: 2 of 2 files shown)


raulvidis:
Reviewed by PR Pipeline agent.

**6-Phase Review — PASS**

**Phase 1 — Context:** `findNodeById` (ActionExecutor.kt:679-695) iterates `roots.withIndex()`. When a match is found at window `wi`, the code recycles `roots.subList(wi+1, ...)` (subsequent roots) and breaks — but never recycled `roots[wi]` itself when the match was a *child* (not the root). `findNodeByIdInTree` returns the matched node without recycling its ancestors (documented memory contract at line 700). The sibling `ScreenReader.findNodeByText:86` already handles this with `if (result !== root) root.recycle()`. This PR adds the identical guard.

**Phase 2 — Security:** No security surface. No new deps, no I/O, no network. Accessibility node recycling is a resource-management concern, not auth/secret/injection.

**Phase 3 — Bugs:** Correct fix. When `found` is a child of `roots[wi]`, the root at the match index leaked — only subsequent roots via subList were recycled. The guard `if (found !== roots[wi]) roots[wi].recycle()` is the idiomatic pattern (reference equality check avoids recycling a root that IS the found node, which the caller owns). No edge case regression: if `found === roots[wi]`, root is preserved for caller use, same as before.

**Phase 4 — Code Quality:** Surgical one-liner. No dead code, no magic numbers. Comment in the test thoroughly explains the regression. Consistent with the established `findNodeByText` pattern — good DRY-by-convention.

**Phase 5 — Testing:** ✅ CI runs `testDebugUnitTest` (verified in `.github/workflows/build.yml`), so the new Robolectric test is genuinely exercised — not dead weight. Test sets up a 2-window scenario where window 0 has no match and window 1's root contains a matching child, then asserts `root1.recycle()` was called. The ID computation (`packageName_className_path_left_top_right_bottom`) and path indexing (`wi_childIndex`) match `findNodeByIdInTree`'s logic. The `@RunWith(RobolectricTestRunner)` + `@Config(sdk=[33])` annotation is added at class level — CI green confirms existing tests still pass under Robolectric.

**Phase 5.5 — Sibling check:** Grepped all `.kt` for the multi-window root-iteration + recycle pattern. Only two implementations: `findNodeById` (ActionExecutor — **this PR**) and `findNodeByText` (ScreenReader:86 — **already fixed**). `searchNodes` (ScreenReader:126) recycles roots unconditionally after DFS (no early-break-on-match, different pattern — correct as-is). `findNodeByIdInTree` and `findNodeByTextDfs` recycle unmatched children during DFS (orthogonal to root recycling). No other siblings. ✅

**Phase 6 — Integration:** No signature change to `findNodeById` (still `private`, returns `AccessibilityNodeInfo?`). All callers (`describeNode:461`, lines 36/296/341) receive an already-recycled-or-owned node — contract unchanged. New test is additive. CI green. Mergeable.

*(Note: PR Pipeline cannot self-approve — GitHub blocks review authors from approving their own PRs. Review posted as comment for the record.)*

PR #84, opened 2026-07-11
auto: URL-encode searchContacts query (URI injection prevention)

raulvidis:
## What was fixed

`ActionExecutor.searchContacts()` appended the raw query string to `ContactsContract.Contacts.CONTENT_FILTER_URI` via `Uri.withAppendedPath()` without URL-encoding. If the query contains URI-special characters (`/`, `#`, `?`), this can break the content URI or cause unexpected query behavior.

The Python side (`android_tool.py:713`) already URL-encodes the query via `quote(query)`, but the Kotlin side received it decoded and passed it raw to the content resolver.

## Fix

Added `Uri.encode(query)` before appending to the filter URI:

```kotlin
val safeQuery = android.net.Uri.encode(query)
val uri = Uri.withAppendedPath(Contacts.CONTENT_FILTER_URI, safeQuery)
```

`Uri.encode()` is the standard Android API for encoding path segments — handles `/`, `#`, `?`, spaces, and all reserved characters.

## Verification

- `./gradlew assembleDebug` — BUILD SUCCESSFUL
- Sibling check: this is the only `withAppendedPath` call in the file

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -592,7 +592,8 @@ object ActionExecutor {
         }
         return try {
             val results = mutableListOf<Map<String, String?>>()
-            val uri = android.net.Uri.withAppendedPath(android.provider.ContactsContract.Contacts.CONTENT_FILTER_URI, query)
+            val safeQuery = android.net.Uri.encode(query)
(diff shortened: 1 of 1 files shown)


raulvidis:
## Automated Review — LGTM

**Phase 2 (Security):** Correct fix. `Uri.withAppendedPath()` does not escape the path segment — an attacker-controlled query containing `/` or `%` could inject unexpected URI paths. `Uri.encode()` is the standard Android approach.

**Phase 5.5 (Sibling check):** `CONTENT_FILTER_URI` with raw query used in exactly one place. No siblings.

**Phase 4 (Quality):** Minimal, targeted. Consistent with existing defensive patterns (`safeLimit` earlier in the same function).

**Verdict: APPROVE.** `Uri.encode()` is reversible and preserves search semantics — the contacts provider decodes correctly.

PR #86, opened 2026-07-11
feat(bridge): background mic/camera capture and remote audio playback

Elloidor:
## Summary
Optional **media capture** and **remote TTS playback** for Hermes Android Bridge, plus the same relay hardening as #85 (safe if both merge).

### Bridge APK
| Endpoint | Purpose |
|---|---|
| `POST /mic_record` | Background mic → AAC/M4A (FGS `microphone`) |
| `POST /camera_record` | Background camera → H.264/MP4 (`back`/`front`, optional audio) |
| `POST /play_remote_audio` | Play base64 audio on device (server-side TTS, etc.) |

Implementation notes:
- **Camera2 dual-looper** (worker vs callback threads) — avoids false `Camera open timed out` when callbacks share a blocked handler (seen on Samsung).
- Portrait **orientation hints**: rear `90°`, front `270°` (OEM-dependent; still verify on device).
- Manifest/FGS: `camera` \| `microphone` \| `mediaProjection` \| `specialUse`; runtime permission prompts on launch.
- Optional **env-based signing** via `HERMES_ANDROID_KEYSTORE*` (no keystores/passwords in git).

### Plugin / tools
- `android_mic_record`, `android_camera_record` → `MEDIA:` outputs  
- `android_speak_default_tts` only if `ELEVENLABS_API_KEY` is set (voice/model overridable via env; no secrets in repo)  
- Relay routes for the new endpoints  
- Relay re-pair / `clear_auth_blocks` / `X-Forwarded-For` / `ANDROID_PUBLIC_URL` (overlap with #85)

### Docs
- `docs/FORK_FEATURES.md`, `BUILDING_APK.md`, `DEPLOYMENT_WSS.md`, `ABOUT_RU_EN.md`

## Security / product honesty
- Stock Android **shows camera/mic indicators** during capture — not stealth.
- Duration-limited one-shot clips, not an always-on silent stream.
- No keystores, API keys, or personal infrastructure in the repository.
- Intended for devices the operator owns/administers with informed consent.

## Relationship to #85
#85 is the **small relay-only** fix. This PR is the **larger feature** set.  
If #85 merges first, this branch may need a trivial rebase; the relay bits are intentionally compatible.

Public showcase tree (same lineage): https://github.com/Elloidor/hermes-android-media

## Test plan
- [ ] Build APK (`assembleDebug`) with SDK 34
- [ ] Grant Accessibility + CAMERA + RECORD_AUDIO
- [ ] `android_mic_record(3000)` → non-empty m4a + playable audio
- [ ] `android_camera_record` back + front → H.264+AAC; confirm front not upside-down on at least one device
- [ ] `play_remote_audio` / optional ElevenLabs path plays on speaker
- [ ] Confirm green camera/mic status indicators appear (expected)
- [ ] Regression: pairing, screenshot, read_screen, basic taps still work

## Version
APK `versionName` in this PR: `0.3.0-media` (adjust as you prefer before release).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,14 @@ hermes-android-bridge/local.properties
 # personal shared standards (symlinks to ../agent-repo-standards)
 .agents/skills
 .agents/AGENTS.base.md
+
+# Signing material (never commit)
+*.keystore
+*.jks
+*.p12
+signing/
+**/hermes-android-release.keystore
+
+# APK artifacts outside Android build dir
+*.apk
+!docs/**/*.apk
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## 0.3.0 — public media fork line (2026-07-11)
+
+### Added
+- Background mic capture (`/mic_record`, `android_mic_record`)
+- Background camera capture (`/camera_record`, `android_camera_record`) with Camera2 dual-looper
+- Remote audio playback (`/play_remote_audio`) for server-generated TTS
+- Optional ElevenLabs helper tool gated on `ELEVENLABS_API_KEY` (voice/model via env)
+- Env-based APK signing (`HERMES_ANDROID_KEYSTORE*`)
+- Env-tunable relay auth rate limits (`ANDROID_AUTH_*`)
+- Docs: `docs/FORK_FEATURES.md`, `docs/DEPLOYMENT_WSS.md`, `docs/BUILDING_APK.md`
+
+### Fixed
+- Relay pairing code not applied when relay already running; auth ban clear on re-pair
+- Client IP for rate limits behind reverse proxies (`X-Forwarded-For`)
+- Samsung false camera open timeout when Camera2 callbacks shared blocked handler
+- Front camera portrait orientation hint (OEM-dependent)
+
+### Security
+- No keystores/passwords/API keys in repository
+- Documented wss-behind-proxy deployment; discourage public cleartext relay
+
+
 All notable changes to this project are documented here. Format based on [Keep a Changelog](https://keepachangelog.com/); this project adheres to Conventional Commits.
 
 ## [Unreleased]
--- a/README.md
+++ b/README.md
@@ -2,6 +2,18 @@
 
 Give your AI agent hands. Remote Android device control for [hermes-agent](https://github.com/NousResearch/hermes-agent).
 
+
+## Optional media extensions (this PR / fork line)
+
+This tree may include optional **background mic/camera capture** and **remote audio playback** for server-side TTS. See:
+
+- [docs/FORK_FEATURES.md](docs/FORK_FEATURES.md)
+- [docs/BUILDING_APK.md](docs/BUILDING_APK.md)
+- [docs/DEPLOYMENT_WSS.md](docs/DEPLOYMENT_WSS.md)
+
+Android shows OS camera/microphone indicators during capture. No silent stealth recording on stock Android.
+
+
 ## How it works
 
 ```
--- a/docs/ABOUT_RU_EN.md
+++ b/docs/ABOUT_RU_EN.md
@@ -1,0 +1,39 @@
+# About / О репозитории
+
+## English
+
+**hermes-android-media** is a **public fork line** of [raulvidis/hermes-android](https://github.com/raulvidis/hermes-android).
+
+It keeps the original design: a companion **Hermes Bridge** Android app + Python tools for [Hermes Agent](https://github.com/NousResearch/hermes-agent) so an agent can control a phone over a WebSocket relay (AccessibilityService).
+
+**This fork adds (among other things):**
+
+- Background **microphone** and **camera** recording (Foreground Service; Android shows OS indicators)
+- **Remote audio playback** for server-side TTS (optional ElevenLabs if configured)
+- Relay hardening: update pairing on a live relay, clear auth rate-limit bans, `X-Forwarded-For` for client IP, env-tunable rate limits, prefer `ANDROID_PUBLIC_URL` in setup instructions
+
+**Not included in git:** keystores, API keys, personal servers, pairing codes.
+
+**Credit:** core architecture and tooling belong to the upstream project. See [FORK_FEATURES.md](FORK_FEATURES.md), [DEPLOYMENT_WSS.md](DEPLOYMENT_WSS.md), [BUILDING_APK.md](BUILDING_APK.md).
+
+**Safety:** use only on devices you own or administer with informed consent. Remote control is powerful.
+
+---
+
+## Русский
+
+**hermes-android-media** — **публичная линия форка** [raulvidis/hermes-android](https://github.com/raulvidis/hermes-android).
+
+Идея та же: приложение **Hermes Bridge** на Android + Python-инструменты для [Hermes Agent](https://github.com/NousResearch/hermes-agent), чтобы агент управлял телефоном через WebSocket relay (AccessibilityService).
+
+**В этом форке, в частности:**
+
+- Фоновая запись **микрофона** и **камеры** (Foreground Service; система показывает индикаторы)
+- **Удалённое воспроизведение аудио** для TTS на стороне сервера (опционально ElevenLabs)
+- Hardening relay: обновление pairing на уже запущенном relay, сброс 429-банов, IP из `X-Forwarded-For`, лимиты через env, в инструкции — `ANDROID_PUBLIC_URL`
+
+**В git нет:** keystore, API-ключей, личных серверов, pairing-кодов.
+
+**Кредит:** ядро архитектуры — у upstream. См. [FORK_FEATURES.md](FORK_FEATURES.md), [DEPLOYMENT_WSS.md](DEPLOYMENT_WSS.md), [BUILDING_APK.md](BUILDING_APK.md).
+
+**Безопасность:** только свои устройства / с согласия владельца. Remote control — серьёзный доступ.
--- a/docs/BUILDING_APK.md
+++ b/docs/BUILDING_APK.md
@@ -1,0 +1,67 @@
+# Building the Bridge APK
+
+## Requirements
+
+- JDK 17+
+- Android SDK platform 34 + build-tools 34
+- Or Docker image with Android SDK (e.g. community images that ship `sdkmanager` + Gradle)
+
+## Debug build (default Android debug keystore)
+
+```bash
+cd hermes-android-bridge
+echo "sdk.dir=$ANDROID_HOME" > local.properties
+./gradlew assembleDebug
+# app/build/outputs/apk/debug/hermes-android-<versionName>.apk
+```
+
+## Reproducible distribution signing (optional)
+
+Never commit keystores or passwords. Use environment variables:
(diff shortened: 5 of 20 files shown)


raulvidis (changes requested):
## Review

This is a substantial feature PR — background mic/camera capture and remote audio playback. The relay hardening overlaps with #85 (and includes the same issues — see that review). Here I'll focus on the new media code.

First: the honesty around OS indicators and consent is appreciated. That's the right approach for a fork that adds surveillance-adjacent capabilities.

---

## 🔴 Critical: Full video base64-encoded over HTTP/WS relay

Camera recordings are encoded as base64 and returned through the JSON relay response. Even at 720p/30fps/H.264 at 3Mbps:

- 5 seconds → ~1.9MB raw → ~2.5MB base64
- 60s max → ~22MB base64 payload in a single JSON response

This will:
1. OOM-kill the aiohttp handler on memory-constrained relay hosts
2. Hit WebSocket frame size limits on many proxies
3. Make every `android_camera_record` call block the relay for the entire duration

Consider streaming or writing to a file and serving via a temp URL, rather than inlining the whole clip in the JSON response.

---

## 🔴 Critical: `version` field re-added to unauthenticated `/ping`

```kotlin
// Before (current main, deliberately removed):
"authenticated" to authenticated
// Version omitted: /ping is unauthenticated and version info
// helps attackers fingerprint the deployment...

// After (this PR):
"authenticated" to authenticated,
"version" to BuildConfig.VERSION_NAME
```

This reverts a deliberate security fix. Version fingerprinting helps attackers target known vulnerabilities. The comment was even left in the codebase on main explaining why it was removed. Keep it removed.

---

## 🔴 Critical: `X-Forwarded-For` spoofing + `?token=` fallback

Same as #85 — see that review. Both issues are present in both relay copies here.

---

## 🟡 High: `Thread.sleep()` blocking inside `Dispatchers.IO`

```kotlin
method == "POST" && path == "/mic_record" -> {
    val result = withContext(Dispatchers.IO) {
        MicRecorder.record(durationMs)  // blocks for up to 120s
    }
}
```

The recording functions use `Thread.sleep(durationMs)` and block. Dispatchers.IO has a default cap of 64 concurrent threads, and if the relay receives multiple concurrent recording requests, it can starve the IO pool for other operations (screenshot, accessibility commands, etc).

Consider `delay()` in a suspend function, or at minimum document that only one concurrent recording is supported and add a mutex/lock.

---

## 🟡 High: No concurrent recording guard

```kotlin
object CameraRecorder {
    // singleton with shared handlerThread
    // no lock on concurrent record() calls
}
```

If two `/camera_record` calls arrive simultaneously, both try to open the camera → second one gets `SecurityException` or blocks the first. Add a `@Volatile var recording = false` guard that rejects concurrent calls.

---

## 🟡 High: `MicRecorder` missing concurrent guard + stale recorder reference

```kotlin
@Volatile private var recorder: MediaRecorder? = null
```

The `recorder` field is set but never read anywhere except `cleanup()`. If a recording is interrupted (e.g. service destroyed), the stale reference persists. More importantly, two simultaneous `record()` calls would clobber each other.

---

## 🟡 Medium: `RemoteAudioPlayer` — no volume control, blocks `prepare()`

```kotlin
val mp = MediaPlayer().apply {
    setDataSource(file.absolutePath)
    prepare()  // synchronous — blocks calling thread
    start()
}
```

`prepare()` is synchronous and could block on slow storage. Also no way to control volume or route audio (speaker vs earpiece). Not blocking, but worth noting.

---

## 🟡 Medium: Files in `cacheDir` never cleaned on failure

Both `MicRecorder` and `CameraRecorder` write to `cacheDir`. On success they `outputFile.delete()`, but on timeout (latch expires without the handler executing cleanup) the files leak. Android may clean cacheDir eventually, but under repeated failures this accumulates.

---

## 🟡 Medium: `HandlerThread`s never stopped

```kotlin
private val handlerThread = HandlerThread("CameraRecorderWorker").apply { start() }
private val callbackThread = HandlerThread("CameraRecorderCallbacks").apply { start() }
```

These start at class init and never call `quit()`. They'll live for the process lifetime. Not a leak per se (singletons), but if the object is initialized but rarely used, the threads idle forever.

---

## 🟡 Medium: ElevenLabs default voice ID hardcoded

```python
voice_id = os.getenv("ELEVENLABS_VOICE_ID", "cgSgspJ2msm6clMCkdW9")
```

Hardcoded voice ID. Not a secret, but it's a specific account's voice. For a public repo, better to require it as env or use a well-known public voice. Also `__import__("base64")` when the file likely already imports it or could.

---

## 🟢 Minor
- README insertion about "media extensions" may confuse upstream users who don't have this fork. Consider making doc additions conditional or clearly marked as fork-specific.
- `CHANGELOG.md` adds a `0.3.0-media` entry above the existing `## [Unreleased]` — breaks the Keep a Changelog convention ordering.
- `versionName = "0.3.0-media"` and `versionCode = 7` — the `-media` suffix is non-standard for semver and may break version comparison logic.

---

## ✅ Good things
- Camera2 dual-looper pattern is the right fix for the Samsung timeout issue — well-reasoned
- Orientation hints (front 270°, rear 90°) are correct for portrait
- Proper permission checks before recording
- `dummyTexture` for preview surface workaround is the standard approach
- Comprehensive error handling with cleanup in CameraRecorder
- Honest security/product notes in the PR description
- No secrets in the repo — env-based signing is correct

---

## Summary

The relay hardening from #85 should land first (after fixes). This PR needs significant rework before it's mergeable: the base64-over-JSON transport for video is architecturally problematic, the version fingerprinting revert must be undone, and concurrent recording needs guards. The camera/mic implementation itself is solid Kotlin — it's the integration layer that needs work.

raulvidis:
Automated review — this PR is based on a stale commit and merging it would revert two merged security fixes:
1. /ping re-exposes versionName to unauthenticated callers (it was deliberately removed — version fingerprinting helps attackers target known vulns).
2. /current_app loses the fix that recycles ALL AccessibilityNodeInfo roots (node-leak fix), reverting to first-root-only.
Also: conflicts with main; versionName '0.3.0-media' collides with the upstream 0.4.0 line; ~1,300 added lines mixing media features, relay changes, signing config, and fork-identity docs in one PR; and it carries the same unconditional X-Forwarded-For trust issue as #85 (rate-limit bypass via spoofed header).
Suggestion: rebase on current main and split into focused PRs — (a) media capture/playback, (b) relay hardening, (c) docs/signing. Happy to review those individually.

raulvidis:
Appreciate the ambition here — remote audio playback and richer capture options are useful directions. However, this PR is on a very stale base (~v0.2.0; current main is v0.4.1) and, as submitted, has five blockers including regressions of documented security fixes and a plugin-registration crash. **Blocked pending rework.**

## Blockers

1. **Relay auth rate limiting is bypassable via unconditional `X-Forwarded-For` trust** (`_client_ip()` in both relay copies). Any client can spoof XFF per attempt, defeating the per-IP counters/bans on `/ws` and HTTP auth — unlimited brute force of the 6-char pairing code. Even the nginx setup in your own `docs/DEPLOYMENT_WSS.md` (`proxy_add_x_forwarded_for`) leaves the spoofed prefix leftmost. *Fix:* only honor XFF when the direct peer is a configured trusted proxy; use the rightmost untrusted hop. (Same issue as #85.)
2. **`?token=` query-param auth fallback re-introduced (tools copy only)** — reverts the #78/#83 hardening (CHANGELOG v0.4.0 Security) that keeps pairing codes out of reverse-proxy access logs. The plugin copy doesn't get the hunk, so the two relay copies diverge functionally. *Fix:* drop the fallback; current APKs use the Authorization header.
3. **Route registration missing from the tools relay copy** — the plugin copy adds `/play_remote_audio`, `/mic_record`, `/camera_record`; `tools/android_relay.py` has no corresponding hunk, so those endpoints 404 in the standalone toolset. *Fix:* add the routes to both copies and reconcile with main's `ROUTES` dict (which already carries #95's `/mic_*` routes).
4. **`android_speak_default_tts` is in `_SCHEMAS` but absent from `_HANDLERS`** (both tool copies). The plugin's `__init__.py` does `handler=_HANDLERS[tool_name]` unguarded → `KeyError` → the entire plugin fails to register. *Fix:* add the handler in both copies, and please add a test asserting `_SCHEMAS.keys() == _HANDLERS.keys()`.
5. **Unauthenticated `/ping` version disclosure re-introduced** (`CommandDispatcher.kt` deletes the deliberate mitigation from fix #81). *Fix:* keep `/ping` version-free.

## Major

- **Rate-limit defaults weakened 4×** (5→20 attempts, 60→120s window) — keep 5/60s defaults; combined with blocker 1 this makes brute force effectively unbounded.
- **Path traversal in `RemoteAudioPlayer.play()`** — the `extension` from the JSON body is interpolated into the cache filename (`File(service.cacheDir, "remote_tts_….$extension")`), so `extension="../../shared_prefs/x"` writes attacker-controlled bytes anywhere in the app sandbox; no size cap on `audio` either (OOM/disk DoS). *Fix:* allow-list the extension (mp3/wav/m4a) and cap decoded size.
- **Headline durations unreachable through the relay** — relay/tool timeouts are 30s but `android_mic_record` allows 120s and `android_camera_record` 60s; the phone blocks for the full duration then replies, so anything >~25s returns 504 while the phone discards the finished recording. #95 solved this with start/stop/fetch + chunked streaming. *Fix:* adopt the async model or parameterize timeouts consistently.
- **Camera recordings break the phone WebSocket above ~7–8s** — the whole MP4 is base64-encoded into one WS TEXT message; aiohttp's default `max_msg_size` is 4 MiB and is never raised, so the socket errors and the relay drops the phone. *Fix:* chunked binary streaming like #95's `/mic_file`, or raise `max_msg_size` with an explicit bound.
- **Stale base: conflicts with and reverts shipped work** — redefines `android_mic_record` with incompatible semantics while main already ships #95's `mic_record/stop/status/fetch` (schema collision); reverts `current_app` multi-root handling to first-root-only. *Fix:* rebase on main and pick one mic architecture.
- **No tests at all** despite surveillance-grade surface. Needed at minimum: auth on the 3 new endpoints, XFF handling, `?token=` absence, schema/handler parity, filename sanitization, duration caps.

## Consent & visibility

Foreground-service notification is present (good), but stays "Connected to server" while camera/mic are actively recording, and the camera|microphone FGS types are never dropped after capture ends (misdeclared FGS state on Android 14, misleading user-visible state). OS privacy dots do show. **Conditional pass:** notification text must state when capture is active, and FGS types should be released after capture. Also, auto-prompting CAMERA + RECORD_AUDIO at app launch without rationale is aggressive consent UX — consider prompting in context.

Suggested path forward: rebase on current main, reconcile with #95's mic architecture, fix the five blockers, split the audio-playback feature from the capture changes if possible, and add tests. Happy to re-review.

rozeappletree:
Shall i work on this? well needed PR imo. Looks stale.

cc: @raulvidis

raulvidis:
@rozeappletree yes please, happy for you to take it over. Please rebase on current `main`, reconcile with #95's mic start/stop/fetch model, fix the 5 blockers listed above (esp. XFF trust + `?token=` + schema/handler parity + `/ping` version), and split audio-playback vs capture if you can. Open a new PR from your fork and I'll review.

PR #87, opened 2026-07-12
auto: add system-destructive broadcasts to blocklist (shutdown/master_clear/factory_reset)

raulvidis:
## What was fixed

The `sendBroadcast` blocklist was missing critical system-destructive actions:
- `android.intent.action.ACTION_SHUTDOWN`
- `android.intent.action.MASTER_CLEAR`
- `android.intent.action.FACTORY_RESET`

While Android blocks these at the system level (protected broadcasts), explicit rejection provides clear feedback to the caller instead of a cryptic system error.

## What was checked
- Sibling implementation check: grepped repo for other `sendBroadcast` callers — only one implementation in ActionExecutor
- Build: `./gradlew assembleDebug` passes
- No new dependencies, no API changes, single-file change

PR #90, opened 2026-08-03
(bridge) sendIntent: denylist dangerous URI schemes incl. single-colon tel:/smsto:

raulvidis:
## Problem

sendIntent previously validated only the *action* string against a denylist (PR #87). A benign action like `android.intent.action.VIEW` with a malicious `dataUri` could bypass that gate — `intent://` redirects, `market://` install flows, `content://settings` / `content://com.android.contacts` providers, and `tel:`/`smsto:`/`mmsto:` telephony triggers.

## Fix

Added a URI scheme denylist in `ActionExecutor.sendIntent` before `setData()`:

- **Scheme extraction fix:** `substringBefore("://")` alone fails for single-colon schemes (`tel:1234` has no `://`, so it returned the full string). Now chains `.substringBefore("://").substringBefore(":")` to correctly extract `tel`, `smsto`, `mmsto`, `intent`, `market`.
- **Blocked schemes:** `intent`, `market`, `smsto`, `mmsto`, `tel`
- **Blocked content prefixes:** `content://settings`, `content://com.android.contacts`

## Tests

8 new tests in `ActionExecutorSendIntentTest`:
- Blocks each scheme: intent://, market:, tel:, smsto:, mmsto:
- Blocks content://settings and content://com.android.contacts
- Allows benign https:// URI (no false positive)
- All assert startActivity is never reached (`verify(exactly = 0)`)

## Review

- Single entry point: `/intent` route in CommandDispatcher; both BridgeRouter (HTTP) and RelayClient (WebSocket) funnel through it, so the gate covers all transports.
- No signature changes, no new deps, denylist pattern mirrors existing sendIntent action block + sendBroadcast block conventions.

### DIFF
--- a/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
+++ b/hermes-android-bridge/app/src/main/kotlin/com/hermesandroid/bridge/executor/ActionExecutor.kt
@@ -663,6 +663,23 @@ object ActionExecutor {
         val intent = Intent(action).apply {
             addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
             if (dataUri != null) {
+                // Denylist dangerous URI schemes that can bypass the action blocklist.
+                // An attacker can use a benign action like VIEW with a malicious URI
+                // (intent:// redirects, content:// providers, market:// deep links,
+                //  tel:, smsto:, mmsto:).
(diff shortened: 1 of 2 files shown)


raulvidis:
Superseded: the URI scheme extraction fix and tests from this PR landed via direct push (558050c) + PR #91.

PR #93, opened 2026-08-04
fix(plugin): align manifest with v0.4.0 release

xiaoyaner0201:
## Summary

The upstream v0.4.0 release updated the Python package and Android app version, but the standalone Hermes plugin manifest/module still reported v0.3.0. This change aligns both plugin metadata surfaces with the already-published v0.4.0 release and adds a regression test that keeps them in sync with `pyproject.toml`.

## Changes

- set `hermes-android-plugin/plugin.yaml` to `0.4.0`
- set the standalone plugin module version to `0.4.0`
- add metadata consistency coverage

## Verification

- Python suite: `132 passed`
- PluginManager registration canary: 38 `android_*` tools
- `git diff --check`
- no Android relay or physical device control was started

This is a metadata consistency fix only; it does not change the Android app or relay behavior.

### DIFF
--- a/hermes-android-plugin/__init__.py
+++ b/hermes-android-plugin/__init__.py
@@ -1,6 +1,6 @@
 """
-hermes-android plugin — registers 14 android_* tools into hermes-agent via the
-v0.3.0 plugin system.
+hermes-android plugin — registers 38 android_* tools into hermes-agent via the
+v0.4.0 plugin system.
 
 Drop this folder into ~/.hermes/plugins/hermes-android and restart hermes.
 """
--- a/hermes-android-plugin/plugin.yaml
+++ b/hermes-android-plugin/plugin.yaml
@@ -1,5 +1,5 @@
 name: hermes-android
-version: 0.3.0
+version: 0.4.0
 description: "Remote Android device control — read screen, tap, type, open apps, take screenshots via the Hermes Bridge companion app"
 author: raulvidis
 requires_env: []
--- a/tests/test_plugin_metadata.py
+++ b/tests/test_plugin_metadata.py
@@ -1,0 +1,18 @@
+from __future__ import annotations
+
+import tomllib
+from pathlib import Path
+
+import yaml
+
+
+ROOT = Path(__file__).resolve().parents[1]
+
+
+def test_plugin_manifest_version_matches_distribution_version():
+    project = tomllib.loads((ROOT / "pyproject.toml").read_text(encoding="utf-8"))
+    manifest = yaml.safe_load(
+        (ROOT / "hermes-android-plugin" / "plugin.yaml").read_text(encoding="utf-8")
+    )
+
+    assert manifest["version"] == project["project"]["version"]


raulvidis:
Automated review: change is clean and correct — verified plugin.yaml 0.4.0 matches pyproject.toml, and the 38-tool count matches _SCHEMAS in android_tool.py. One issue: tests/test_plugin_metadata.py imports yaml, but pyyaml is not declared in pyproject dependencies (or the dev extra) — the test will fail in a fresh venv. Please add pyyaml to [project.optional-dependencies].dev. Also note: this PR can never pass the required 'build' status check because build.yml is path-filtered to hermes-android-bridge/** — it needs an admin merge or a CI trigger update. Waiting on maintainer.

raulvidis (changes requested):
## Review: stale — superseded by v0.4.1 on `main`

**Verdict: request changes** — recommend closing, or rebasing to keep only the regression test.

This PR fixes a problem that no longer exists on `main`. It was opened against v0.4.0 on Aug 4; main has since moved through the v0.4.1 release (`08770ab`), then #94 and #95:

| Change in this PR | State on current `main` |
|---|---|
| `plugin.yaml` 0.3.0 → 0.4.0 | Already **0.4.1** — merging would conflict with / revert the manifest, and this PR's own regression test would then fail (0.4.0 ≠ 0.4.1) |
| `__init__.py` docstring: "38 android\_\* tools … v0.4.0 plugin system" | Already rewritten to "42 android\_\* tools through the hermes-agent plugin system" (version refs removed) — hunk will conflict |
| `tests/test_plugin_metadata.py` | The only surviving value, but needs fixes below |

Issues to address if rebased:

1. **[MAJOR] Missing dependency** — the test imports `yaml`, but `pyyaml` is not in `[project.optional-dependencies].dev` → fails in a fresh venv (also flagged by the earlier automated review).
2. **[MAJOR] CI never runs** — `build.yml` is path-filtered to `hermes-android-bridge/**`, so this PR can never produce the required `build` check. Needs a workflow-trigger update or an admin merge.
3. **[NIT]** The test only asserts the manifest version; the docstring tool-count has drifted twice already (14 → 38 → 42). Consider also asserting `len(_SCHEMAS)` against the docstring claim.

Suggested path: **close as superseded**, or rebase onto current main and keep only the regression test + `pyyaml` in dev extras.

raulvidis:
Following up on the review above: since the version-sync goal is already on `main` (0.4.1), I'll close this as superseded in ~1 week unless you'd like to rebase it down to just the regression test + `pyyaml` dev dep — happy to take that rebased version over if it's easier. Thanks either way; the regression-test idea itself is a good one.

xiaoyaner0201:
Thanks for the thorough review and the follow-up — agreed on both counts.

Confirmed on current `main`: the version-sync goal this PR was opened for is already achieved by v0.4.1 (`08770ab`) plus #94/#95, so the substantive change here is obsolete. Closing rather than rebasing, to keep your queue clean.

Two notes for whoever picks this up later:

- The `pyyaml` gap you identified is real and independent of this PR: `tests/test_plugin_metadata.py` imports `yaml`, but `pyyaml` is not declared in `[project.optional-dependencies].dev`, so the test fails in a fresh venv.
- The regression test idea (asserting `plugin.yaml` version matches `pyproject.toml`, and the tool count matches `_SCHEMAS`) still has value as a guard against future drift.

Happy to open a separate minimal PR for just those two if you'd like it — otherwise feel free to fold them in whenever convenient. Thanks for maintaining this.

PR #95, opened 2026-08-11
feat: Microphone support Phase 1 (Ambient Sound Recording)

Thykes3000:
## What
Adds microphone recording to the Hermes Bridge app — the foundation for ambient sound analysis and voice capture (Roadmap v0.5 "Voice assistant mode").

## Changes
- **Android (Kotlin):**
  - `MicrophoneRecorderService.kt` — Foreground Service with `foregroundServiceType="microphone"` (Android 14+ required for continuous mic), uses `AudioRecord` → writes WAV (16 kHz, mono, PCM16)
  - Manifest: `RECORD_AUDIO` + `FOREGROUND_SERVICE_MICROPHONE`
  - New bridge routes: `POST /mic_start` (duration optional), `POST /mic_stop`, `GET /mic_status`
- **Python toolset:** 4 new tools — `android_mic_record`, `android_mic_stop`, `android_mic_status`, `android_mic_fetch`

## Why
The phone's microphone is the last missing sensor for a fully autonomous edge device. Phase 1 records stable PCM/WAV via the supported Android path (no shell hacks). Subsequent phases: VAD, then STT (Whisper/Vosk), then wake word — exactly as planned in the roadmap.

## Testing
- `AudioRecord` initialized at 16kHz mono on Pixel 6 Pro (Android 14)
- WAV header written correctly (44-byte RIFF, sizes patched on stop)
- Foreground service notification shown while recording

## Notes
- Records to app-specific external dir (`cradata_audio/`) — no storage permission needed
- Start: `POST /mic_start {"duration": 10}` (0 = until stop)
- Files retrieved via SCP/ADB or a future file-transfer endpoint

PR #96, opened 2026-08-14
Add robot dialog, Realtime tiers, and noise-triggered video

Thykes3000:
## What changed

- add a cinematic Android robot-dialog screen with explicit GPT Live, Hermes Local, and Hermes Standard backends
- add three Mac-controlled OpenAI Realtime tiers (`mini`, `standard`, and `top`) without shipping an API key in the APK
- add the restricted Hermes dialog companion, warm Whisper worker, private text transcript archive, and memory-only observer
- add noise-triggered rear-camera video with bounded clips, retention, foreground-service visibility, and authenticated fetch/status routes
- pause the noise watcher's microphone while RobotActivity owns it, then restore the previous watcher configuration when the dialog closes
- harden microphone recording finalization, VAD, retention, relay streaming, SDP validation, safe upstream error mapping, and reconnect cleanup
- add boot reconnect support, installer packaging, public documentation, and regression tests

## Why

The Pixel is being used as a robot body and conversational surface. Two shared resource failures blocked the end-to-end experience:

1. the continuous loud-noise watcher could retain the microphone, making the robot screen flash "listening" and immediately fail;
2. valid WebRTC offers could become invalid when the relay stripped the required trailing SDP CRLF or omitted the SDP multipart content type.

The bridge now arbitrates microphone ownership and preserves byte-correct SDP while keeping provider credentials on the Mac.

## Security and privacy

- OpenAI and relay credentials remain Mac-only and are never returned to Android
- Realtime errors are reduced to bounded machine-readable slugs before reaching the Pixel
- robot answer workers have memory tools only, with no device, purchase, message, or call capabilities
- push-to-talk WAV files are temporary; GPT Live audio is not archived by this project
- completed media uses bounded size and retention policies
- private operator handover files, addresses, pairing data, media, and local runtime configuration are deliberately excluded from this branch

## Validation

- `uv run --extra dev pytest -q` — **207 passed**
- `./gradlew :app:testDebugUnitTest :app:assembleDebug` — **passed**
- `git diff --check` — **clean**
- current repository and deployed relay Python copies were compared byte-for-byte

## Review note

This is intentionally a draft for scope and architecture review. It is a clean branch from current `main`, not a continuation of any branch history containing private operator documents. The feature set can be split into smaller follow-up PRs if maintainers prefer. A final on-device audio/video pass with the clean APK is still required before marking it ready for merge.


raulvidis:
## Security & architecture review (draft stage)

Two independent deep reviews were run over the full 9.2k-line diff — one covering the Kotlin/Android side, one the Python relay/tools side — plus a secrets scan of every hunk. Overall: **the security architecture is genuinely good, but there are 2 MAJOR items to address before this can leave draft.**

**Verdict: request-changes** (Kotlin side: 2 majors; Python side: approve-with-nits)

---

### MAJOR

**M1. Remote `/noise_watch_stop` can crash the whole bridge app** — `NoiseTriggeredVideoService.kt` (`onCreate` + companion `stop()`) via `CommandDispatcher` `/noise_watch_stop`.
`stop()` is unguarded (no "is it running" check) and `onCreate` promotes to foreground with `FOREGROUND_SERVICE_TYPE_CAMERA or FOREGROUND_SERVICE_TYPE_MICROPHONE` before any action handling. On `targetSdk 34`, `startForeground` with those types throws `MissingForegroundServicePermissionException` when CAMERA/RECORD_AUDIO aren't granted — the common case, since the watcher is opt-in. So a routine `android_noise_watch_stop()` call when the watcher was never enabled creates the service → crash → bridge process (Ktor + relay WS + accessibility) dies.
Fix: guard `stop()` with `if (!NoiseTriggerState.snapshot().active) return`, move foreground promotion into the `ACTION_START` path after the permission check, and/or wrap `startForeground` in try/catch. Please also verify manually on an Android 14 device without the permissions granted.

**M2. Unconditional auto-reconnect on every boot and every app update, no opt-out** — `BootReceiver.kt` + manifest (`BOOT_COMPLETED`, `LOCKED_BOOT_COMPLETED`, `MY_PACKAGE_REPLACED`).
Combined with the existing `RelayClient.autoConnect()`, every reboot — and every APK install (which this repo does frequently via `latest-build` auto-publish) — silently re-establishes full remote control with no user interaction. This directly undercuts SECURITY.md's "Disconnect when not in use" mitigation, and there's no UI toggle. The broadcasts are protected (no spoofing risk), so this is a policy issue, not an injection hole — but it needs an opt-in toggle (e.g. `setComponentEnabledSetting`) or at least a first-boot confirmation, plus SECURITY.md documentation.

---

### MINOR

**m1.** `_handle_realtime_session`: the `OpenAI-Safety-Identifier` is an HMAC keyed with the long-lived OpenAI key over the low-entropy pairing code (32^6 ≈ 1.1e9). If key + header value are ever disclosed together (e.g. provider-side logging), the pairing code becomes offline brute-forceable → relay credential recovered. Use a random per-session identifier (or HMAC with a server-local secret) — or drop the header if upstream doesn't consume it.
**m2.** SDP answers (carrying ICE credentials for live voice sessions) and robot events travel plaintext over `http://`/`ws://` unless the operator fronts the relay with TLS — an extension of SECURITY.md's documented "No Encryption" limitation. Please add one sentence about the new robot/realtime plaintext surface to SECURITY.md Known Limitations.
**m3.** `/noise_watch_start` parameter parsing (`thresholdRms`, `clipSeconds`) lacks the `isJsonPrimitive` type checks that `/mic_start` in the same PR uses — malformed JSON yields a generic 500 instead of 400. Mirror the `/mic_start` pattern.
**m4.** No tests for the path-traversal defenses: `NoiseVideoFiles.resolve()` (backs `GET /noise_video_file` + WS stream) and — from the merged mic code it models itself on — the security-relevant validation deserves regression tests: `../`, absolute paths, URL-decoded variants, symlinks, wrong-case extensions, unknown IDs.
**m5.** Relay binary-stream verification has no negative-path tests (SHA-256 mismatch, oversize beyond declared, truncation, mid-stream disconnect) — these are exactly the lines that make the feature safe.
**m6.** `robot_dialog.py` live-transcript pairing: user/assistant utterances are paired from two independent FIFOs; an unanswered user utterance gets paired with the *next* assistant reply, so the memory observer can receive a mismatched pair. Pair by adjacency/item-id, or drain unmatched entries on `robot.session_stop`.
**m7.** `ConversationArchive` JSONL has per-entry bounds and 0600 perms but no retention/rotation → unbounded growth under `~/.hermes/robot-dialog/transcripts/`. Add an optional max-count/age cap mirroring phone-side retention.
**m8.** Concurrency hygiene in `NoiseTriggeredVideoService`: config vars written on main thread, read from `Dispatchers.IO` without `@Volatile` — inconsistent with the careful `@Volatile` discipline the same PR uses elsewhere. Also decide/document stop semantics mid-clip (currently the in-flight clip still completes after "stop").

---

### Verified claims (positives)

- **Auth continuity holds end-to-end.** All new surface (`/robot_state`, `/robot_status`, `/noise_watch_*`, `GET /robot/events`, `POST /robot/realtime/session`, `/noise_video_file`) flows through the existing Bearer-pairing-code checks on both transports (Ktor interceptor; relay `_http_auth_error` with rate limiting). No unauthenticated or weakly-validated entry points.
- **Credentials never reach Android — confirmed.** `OPENAI_API_KEY` is read only Mac-side; the APK ships tier names + SDP only, upstream errors collapse to bounded slugs (`[A-Za-z0-9_.:/+-]{1,160}`, statuses {400,401,403,429,502}), and HTTP error handling deliberately discards response bodies.
- **WebView hardening is exemplary**: JS only for the bundled asset, file/content access off, all navigation blocked, `MIXED_CONTENT_NEVER_ALLOW`, strict CSP with `.invalid` base origin, `onPermissionRequest` grants audio-capture only to the trusted origin, SDP injected via `JSONObject.quote`.
- **Worker confinement is real, not a comment**: hard `toolsets="memory"` at the process boundary, stdin-only message passing, `@file` neutralized to `＠`, model/provider slug-validated with injection tests. (Note: ultimate strength depends on hermes-agent's `cli.main` semantics — see Q1.)
- **Noise watcher privacy controls**: opt-in switch + runtime perms, visible ongoing notification, ≤30 s clips @ 2 Mbps, 10-file retention + stale-`.part` cleanup, owner-only dirs, `no-store` + attachment disposition, SHA-256-checked relay streaming.
- **PII conventions preserved**: `caption` added to relay log masking, cloud-bound vs memory-bound redaction layers (both unit-tested), `ActionExecutor.speak` no longer echoes spoken text, event logging emits names only.
- **Copy parity holds**: both `android_relay.py` copies byte-identical; the `android_tool.py` +174 vs +170 difference is exactly 4 docstring lines (the `android_noise_*` tool-list entries) — all functional hunks identical.
- **Secrets scan clean**: no keys, tokens, real IPs, emails, or personal identifiers in any hunk — only obvious fixtures.

---

### Nits

- Plugin copy's docstring header still says "42 android_* tool handlers" while README/skills moved to 46.
- `robot_dialog.py` entry point: `logging.basicConfig` is only in the `__main__` guard → INFO logs silent via console script; move it into `main()`.
- `whisper_worker.py`'s `openai-whisper` dependency is undocumented in packaging — consider an optional `[voice]` extra.
- `OpenAIResponsesProvider` (`chat-latest`) is fully implemented but unreachable — dormant code; remove or mark reserved.
- Mic contention: `/noise_watch_start` checks only the WAV recorder state, not the robot dialog's live mic — a `RobotRealtimeController.active` guard would make the 409 story symmetric.

---

### Open questions for the author/maintainer

1. Does hermes-agent's `cli.main(toolsets="memory")` guarantee zero device/purchase/message/call tools — including when `resume=<session_id>` restores a persisted session? Recommend pinning/documenting the required CLI contract.
2. Is boot/update auto-start (M2) intended to be unconditional?
3. Has `/noise_watch_start`'s `startForegroundService` for a camera|microphone FGS been verified on a real Android 14 device from a backgrounded app (background-FGS-start restrictions)? If restricted, the feature would silently be UI-only.
4. `selectedRealtimeTier` defaults to `TOP` (highest-cost tier) — intentional product default?
5. Model slugs (`gpt-realtime-2.1-mini`, `gpt-realtime-2`, `chat-latest`) — please confirm these resolve upstream before release.
6. Should `/robot_state show=true` work while the device is locked?
7. Is unbounded transcript-archive growth (m7) intentional for archival use?

Validation still owed per the PR's own note: final on-device audio/video pass, plus `./gradlew test` (6 new Kotlin test files incl. Robolectric) and `pytest tests/`.

raulvidis:
Maintainer decisions on the open questions, so the next revision can proceed:

1. **Boot/update auto-reconnect (M2 / Q2):** make it a UI toggle (e.g. gate `BootReceiver` via `PackageManager.setComponentEnabledSetting`), **default ON** — the typical deployment is a dedicated bridge device — but the behavior must be user-visible and documented in SECURITY.md, whose "disconnect when not in use" guidance currently contradicts it.
2. **Default Realtime tier (Q4):** change the default from `TOP` to `standard`; operators can explicitly opt up. Cost control should be the default posture.
3. **Robot UI on a locked device (Q6):** suppress `/robot_state show=true` while the device is locked (KeyguardManager check). The robot screen should not be poppable remotely on a locked device.
4. **Transcript archive retention (Q7):** add bounded retention mirroring the phone-side media policy (e.g. last 200 sessions or 90 days, whichever is sooner), with an env override (`ROBOT_DIALOG_TRANSCRIPT_ARCHIVE_LIMIT=0`) for operators who genuinely want unlimited archival.

Everything else in the review stands. Priorities for the next revision: **M1** (the `/noise_watch_stop` crash — please include the manual Android 14 verification without camera/mic permissions granted) and **M2**, then the minors (m1–m8). Q1, Q3 and Q5 remain on your side: the hermes-agent `cli.main(toolsets="memory")` contract including `resume`, Android 14 background FGS start behavior on a real device, and confirming the model slugs resolve upstream.

Thanks for the clean branch and the security-conscious design — the architecture itself is in good shape.

Thykes3000:
Quick follow-up on the three questions you left on my side — plus one
piece of housekeeping first, because it affects what you can see: PR #96
still points
at codex/robot-dialog-v2, which predates the review fixes. Everything below
sits on codex/robot-dialog in my fork:

  https://github.com/Thykes3000/hermes-android/tree/codex/robot-dialog

Two routes, whichever suits you: I push these commits onto the existing PR
branch so the diff updates in place, or I open a fresh PR from the verified
branch and close #96. Say which and I'll do it.

Your four decisions, mapped to the relevant commits:

  M2 - auto-reconnect toggle, default ON, documented
       https://github.com/Thykes3000/hermes-android/commit/c4c975345d
  M1 - idle noise-watcher stop no longer kills the app
       https://github.com/Thykes3000/hermes-android/commit/cc6da226b0
  Q4, Q6, Q7 - tier default, keyguard suppression, archive retention
       https://github.com/Thykes3000/hermes-android/commit/8a58505235
  m1, m3-m6, m8 and the remaining minors
       https://github.com/Thykes3000/hermes-android/commit/9d35e0da24
       https://github.com/Thykes3000/hermes-android/commit/bfca1ecc84

Tests, both re-run today: pytest 306 passed; ./gradlew test BUILD
SUCCESSFUL, 129 unit tests per variant, 0 failures.


Q1 — THE CONTRACT HOLDS, RESUME INCLUDED
The toolset is defined as one tool with no transitive includes, the
comma-separated argument survives the boundary as a one-element list, and a
full startup with MCP discovery reports "Total: 1 tools" — this profile has
an MCP server registering 9 further tools and none of them appear, because
server names join the enabled set only when the caller passes no explicit
list. No device, purchase, message or call tool is reachable.

Resume cannot widen that surface: there is no toolset column in the sessions
schema, and the enabled set is assigned once at construction and only read
afterwards. Resume restores existing session options — working directory,
approval settings, model and provider — none of which add tools.

Worth documenting as the contract: pass the toolset explicitly *and* always
pass the model explicitly. Without the second, a resumed session reinstates
the model it last ran with.


Q1, ADDENDUM — ONE DEFECT FOUND WHILE PINNING THAT DOWN
Your note that the strength depends on cli.main semantics was the right
place to look.

Cause:  the CLI entry point the worker uses does not expand configured model
        aliases; a different entry point does, so the requested local model
        was not honoured.
Impact: agent-path turns — spoken recall requests — were served by the
        configured remote default instead. The fast path was never affected.
Fix:    the worker resolves the name itself, with regression tests for an
        explicit provider and a bare model id.
        https://github.com/Thykes3000/hermes-android/commit/4e6f7a9bf7

Follow-up: that local model turned out to be incompatible with the agent
path anyway — it does not meet a minimum context-window requirement — so
the agent path now uses one that does. Compatibility is verified end to end;
latency on that path is not yet where it belongs and I am treating it as
open.
        https://github.com/Thykes3000/hermes-android/commit/528fbada1d


Q5 — THE SLUGS RESOLVE
All four are present on the account that would serve them. Caveat: that
proves the slug is known, not that a Realtime session start succeeds — I
have not minted a live session per tier. The Responses model is also only
reachable through the provider your review flagged as dormant.


Q3 — STILL OPEN, AND I CANNOT CLOSE IT
It needs an Android 14 device with camera and microphone not granted. The
device here runs a newer Android, where the restriction does not reproduce,
and there is no API 34 emulator set up. I would rather leave it openly
unverified than claim it from a device that cannot show the failure. If you
have a 14 device: one stop call with the watcher never started reproduces
it.

Same honesty on decision 3: the keyguard suppression is unit-tested through
a seam, not on the device — the phone here has no screen lock configured, so
the locked branch cannot be triggered on it.

Best,
Robert

Am Di., 18. Aug. 2026 um 02:38 Uhr schrieb Raul ***@***.***>:

raulvidis:
Thanks for the thorough follow-up — the fix mapping, the Q1 analysis, and the candor about what you couldn't verify are all appreciated. Decision first: **option (a)** — push the fix commits onto the existing PR branch so the diff updates in place and this thread stays in one piece.

I fetched `codex/robot-dialog` from your fork and verified before replying. What checks out:

- **M1 is correctly fixed** — idle stop is a no-op via the state snapshot, FGS promotion moved after the permission check, and refusal now surfaces as a failed state instead of killing the bridge process. Matches the review analysis exactly.
- **Q4/Q6/Q7 implemented as decided**, including the SECURITY.md rewrite to describe actual behavior. Noted the toggle shipped off first and your decisions commit flipped it to default ON — final state matches my call.
- **Q1 analysis is sound** — enabled set assigned once at construction, no toolset column in the sessions schema, resume restores options but can't add tools. The contract worth documenting is yours: pass the toolset *and* the model explicitly. The alias-expansion defect you found while pinning that down, fixed with regression tests, is good practice.

Three things to do before the next review round — the first two I found while verifying, so flagging them explicitly:

1. **Rebase onto current main.** Your branch's merge-base is v0.4.1, which predates the mic recording work (#95) — `android_mic_record/stop/status/fetch` and the relay streaming routes already ship upstream, so the collision the review flagged is still live. Several fixes landed since as well (#101, #98/#99). Reconcile `versionCode`/`versionName` during the rebase (your branch is at 17/"4.2.6", main at 3/0.4.1 — pick one scheme, we'll sort the final numbers at tag time).
2. **Make the robot-dialog tests hermetic.** On a bare checkout I get **299 passed, 7 failed** — the failures are `test_robot_dialog.py` tests that hard-require a hermes-agent installation (`~/.hermes/robot-dialog/cli.py`) and raise instead of skipping. Your "306 passed" only reproduces on a machine with the agent installed; bare checkouts and any CI that runs pytest won't. Add `skipif` guards for the missing-installation case.
3. **Q3 closure** — pragmatic path, since neither of us has the Android 14 hardware: your M1 fix makes FGS refusal observable (`NoiseTriggerState.failed`) rather than silent, which was the real hazard. Document the residual Android-14 background-FGS-start risk in SECURITY.md (what can still fail, how it surfaces, that it remains unverified on API 34) and we treat it as accepted risk, not a blocker. Same treatment for the keyguard branch: seam-tested is fine, note the on-device gap.

Q5 is acceptable at draft stage — slugs resolving on the serving account with the live-session caveat stated is enough until we're cutting a release.

On scope: the branch has grown since the review (+20k lines now), but given the review already invested in this shape, I'll take it as one PR once rebased with the test fix — on the understanding that if the rebase balloons it further we revisit splitting.

Push when ready and ping here; I'll re-review the rebased diff.

raulvidis:
Hi Robert — friendly check-in on this one, no rush intended.

Main has moved a bit since your last push (#98/#99 mic fixes, #101, and the Termux revival watchdog from #100 with its post-merge cleanups), so I ran a trial merge of `codex/robot-dialog` onto current main to see where things stand. Result: 23 conflicting files, concentrated in the places we'd expect — the two `android_relay.py` copies (your branch predates the #95 streaming routes), `MicrophoneRecorderService.kt`, `RelayClient.kt`, `CommandDispatcher`, and the dual `android_tool.py` copies. Docs/layouts/CHANGELOG conflicts look mechanical.

Nothing in there changes the plan we agreed (rebase onto main, `skipif` the install-dependent robot-dialog tests, reconcile version numbers) — just a bigger surface than in August.

Two offers, whichever suits you:

1. If you'd rather I handle it, I'm happy to do the rebase myself on an upstream branch — I've reviewed these conflict areas a few times now and know which way each decision goes (main's mic architecture wins, parity rule preserved). You'd still get full authorship credit.
2. If you're mid-work on it (I saw recent branch activity on your fork), ignore the above — push whenever it's ready and I'll re-review promptly.

Either way the architecture verdict stands: the design is in good shape, it's the rebase + test hermeticity standing between this and a merge.

Thykes3000:
Hi Raul,

sorry for the long silence, and thanks for the trial merge and the offer.

Yes, please take option 1 and do the rebase on an upstream branch. You know
which way each of those conflicts goes, and that will be faster than me
catching up with #98–#101.

Three notes for it:

1. Please swap the sample name in the redaction fixtures of
tests/test_robot_dialog.py for a neutral one (my later branch uses
"Jonas"). The current one should not carry forward.

2. Everything still open on my side, from your 20 August list and from the
first review, goes into one small commit on top of your branch once it is
up:
- skipif guards for the robot-dialog tests that need a hermes-agent
installation, so a bare checkout passes
- SECURITY.md: the Android 14 background FGS start as documented residual
risk (what can still fail, how it surfaces, unverified on API 34), and the
note that the keyguard branch is seam-tested only
- the three remaining nits: the plugin docstring still says 42 tools
instead of 46; the openai-whisper dependency gets a documented optional
extra; and /noise_watch_start also answers 409 while the robot dialog holds
the microphone, so the conflict handling is symmetric
If you would rather fold any of this into the rebase, that is fine too.

3. Scope: feature/persistent-dialog-worker on my fork continues the robot
dialog (warm worker between spoken turns, model load while the user is
still speaking). I will keep it out of #96 and bring it as a separate PR
after #96 is merged, so this one does not grow any further.

Ping me here when the branch is up.

Best,
Robert

eabase:
I love how you guys are handling this. Clear and precise! 


<sub>`Rant warning!`  
I wish I could say the same of the [hermes-agent](https://github.com/NousResearch/hermes-agent/) repo... They're a total mess.  
Their entire development process is totally out of control with substantial and  
[blocking security issues](https://github.com/NousResearch/hermes-agent/issues/107356).</sub>

raulvidis:
Housekeeping ahead of the rebase — three things that will otherwise look like the rebase's fault, plus confirmation your list is tracked.

### 1. `main` had a flaky Python suite; it's fixed in #105

`TestMicrophoneBinaryStreamNegative` (added in `fbd6238`, the #99 test debt) was failing on **8 of 12** full-suite runs on `main`, always the same three tests, and never when the class ran alone. Two races, neither related to your branch:

- The two `caplog` assertions read `caplog.records` the moment `asyncio.run(scenario())` returns, but the relay reaches its verdict in the HTTP handler on its **own background thread and event loop**, after the phone's `end` event is processed. The client can be gone first. Fixed with a polling wait (`_await_log_record`) rather than a sleep, so a real regression still fails readably.
- `test_phone_disconnect_mid_stream_tears_down_the_download` permitted only a torn-down body. Instrumented over 20 runs it's **18× `ClientPayloadError`, 2× a clean HTTP 502** — both are the relay refusing to present a truncated stream as complete, so the test now accepts either and still fails on a body that reads cleanly with a 2xx.

#105 is green on 3.11 and 3.13. **It also adds `.github/workflows/python.yml`**, which runs `pytest -q` on every PR touching `tools/`, `tests/`, `hermes-android-plugin/` or the packaging files. So once your `skipif` guards land, a bare-checkout failure is caught by CI instead of by whoever next clones the repo — the 299-passed/7-failed split I hit in August won't be able to recur silently.

Worth knowing while you rebase: if you see red in `test_android_relay.py`, check whether #105 has merged into your base yet. Before it, roughly one run in three is red for reasons that have nothing to do with your changes.

### 2. Port collision waiting in `tests/test_android_relay.py`

Both sides allocate fixed ports and they overlap:

- `main`: `TestMicrophoneBinaryStreamNegative` → **19884**
- your branch: `TestRobotPhoneEvents` → **19884** (then `TestRealtimeSession` 19885, `TestVideoSignaling` 19886)

After the rebase both classes live in one file. They don't run concurrently today, so it may well pass — but it's one `pytest-xdist` or one leaked relay away from an intermittent failure that will look exactly like the flake above and cost someone an afternoon. Cheap to fix during the rebase by renumbering your three classes onto a distinct block.

Related, and it's the reason I'm confident about the above: when I first measured this I got `EADDRINUSE` on 19884 and nearly wrote it up as a teardown bug in `main`. It was my own two worktrees running both suites at once. The real races are the two in #105 — no socket leak, `stop_relay()` joins correctly and `_serve` awaits `runner.cleanup()`.

### 3. Your m5 tests are the better design

`TestVerifiedStreamRejections` on your branch tests the extracted loop directly — 20/20 stable across my runs, no `caplog`, no client-side race. `main`'s `TestMicrophoneBinaryStreamNegative` covers the same ground over a live relay, which is where the races came from. When you rebase, **prefer yours** for the cases they both cover rather than keeping both; the integration-level ones are the expensive, flaky half.

### 4. Your list is tracked

Confirming the three items from your 11 September note so nothing gets lost in the rebase:

- **The redaction fixture name.** Agreed, and it should not carry forward — `tests/test_robot_dialog.py:107` asserts `"Justus" not in redacted`. Swapping to a neutral sample is part of the rebase, not something you need to commit on top.
- **`skipif` guards** for the install-dependent `test_robot_dialog.py` tests — still yours per your note, and now CI-enforced rather than convention-enforced.
- **The three nits** (plugin docstring 42 → 46, the `openai-whisper` optional extra, `/noise_watch_start` answering 409 while the robot dialog holds the mic) — unchanged, still open.

### One conflict to expect that's new since we last spoke

#106 (`GET /battery`) touches `CommandDispatcher.kt`, which is already one of the hot files in your 23-conflict trial merge. It also lifts the relay's `ROUTES` allowlist out of `_serve()` to module level as `_ROUTES`, and your branch's copy of `android_relay.py` predates that — so expect a conflict there that resolves in favour of the module-level version. Your new routes (`/robot_state`, `/robot_status`, `/noise_watch_*`, `/noise_video_file`, `/robot/events`, `/robot/realtime/session`) need to be in `_ROUTES` to be reachable over the relay at all, and there's now a test that fails if the allowlist and `docs/architecture.md` disagree, so it'll flag itself rather than failing silently the way it would have before.

No action needed from you on any of this — flagging it so the rebase doesn't get blamed for things that predate it.

